PR #125, opened 2019-03-30
[docker] use official docker and docker-compose

### LINKED ISSUE #115: Update docker installation
### Overview

Docker and `docker-compose` are not properly configured. Install instructions available in the Ansible script are outdated and need improvements.

palazzem:
### Overview

Closes #115 

* syntax refactoring
* Add `docker-compose` from official repositories

### Caveat

This patch must be tested on a ChromeOS (Crostini) because a Docker patch was not available in the latest release. If that doesn't work, we should use `docker-git` (AUR) package to include the required patch.

### DIFF
--- a/arch/roles/orchestrators/tasks/docker.yml
+++ b/arch/roles/orchestrators/tasks/docker.yml
@@ -3,7 +3,11 @@
 # Docker
 
 - name: Installing docker containers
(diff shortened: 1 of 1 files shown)


palazzem:
Closes also https://github.com/palazzem/hanzo/issues/92

PR #238, opened 2026-05-13
fix(tasks): shell-quote interpolated values and DRY up pull logic

### LINKED ISSUE #234: Fix code quality in dotfiles.py, tools.py, packages.py
## Summary

Three code quality improvements identified during repo review:

### Finding #5: Document @local-only assumption for control-time branching in dotfiles.py

`os.path.isdir()` calls on lines 31, 93, 101 of `tasks/dotfiles.py` evaluate at control time (when pyinfra parses the task file), not at deploy time on the target host. This only works correctly because Hanzo always targets `@local`. A comment should document this assumption, following the pattern already used in `tasks/tools.py` line 62: `# Expanded on the control machine — safe because Hanzo always targets @local.`

### Finding #6: Fix git pull --ff-only fragility in dotfiles.py

Lines 34-36 and 96-98 use `git pull --ff-only` which will fail (and abort the entire pyinfra run) if the local branch has diverged from the remote. Add a fallback so diverged branches produce a warning instead of aborting:

```
git pull --ff-only || echo 'WARNING: fast-forward failed, skipping pull'
```

### Finding #7: Apply shlex.quote() consistently in tools.py and packages.py

`tasks/dotfiles.py` already uses `shlex.quote()` for all shell-interpolated values. The same pattern should be applied to:

- `tasks/tools.py` lines 89, 103, 131-132 — uv tool names, paru package names, npm package names
- `tasks/packages.py` line 52 — paru package names

Import `shlex` at the top of each file and wrap all `host.data.*` values interpolated into `server.shell` commands.

## Files to modify

- `tasks/dotfiles.py`
- `tasks/tools.py`
- `tasks/packages.py`

palazzem:
### Related Issues

- fixes #234

### Proposed Changes:

Three related code quality fixes across the task layer:

**Shell injection hardening (`packages.py`, `tools.py`):** `shlex.quote()` is now applied to every package name before it is interpolated into a shell command string (`paru -S`, `uv tool install`, `npm install -g`). Without quoting, a package name containing shell metacharacters could silently corrupt the command or, in adversarial scenarios, execute unintended shell code. Values come from `group_data/all.py` which is under version control, so the practical risk is low today — but the fix removes an entire class of latent bug and makes the intent explicit.

**Graceful pull fallback (`dotfiles.py`):** The `git pull --ff-only` commands for both dotfiles and claude-config repos previously caused the entire provisioning run to abort if the local branch had diverged (e.g. after a manual commit or rebase). The new behaviour warns and continues — a stale checkout is acceptable; a failed provision is not.

**DRY helper (`dotfiles.py`):** The identical clone-or-pull pattern was duplicated for both repos. `_git_pull_ff_cmd()` extracts the command construction into a single place so the fallback logic only needs to change once.

### Testing:

- Existing behaviour is preserved for the happy path (fast-forward succeeds).
- The fallback path (diverged branch) now emits a `WARNING:` line instead of a non-zero exit.
- `shlex.quote()` is stdlib and has no observable runtime difference for well-formed package names.
- Manual verification: `pre-commit run --all-files` passes; container build passes.

### Extra Notes (optional):

The `os.path.isdir()` calls in `dotfiles.py` evaluate at Python parse time (control machine), not deploy time. This is safe because Hanzo always targets `@local`. The commits add comments documenting this assumption explicitly so future readers don't confuse it with a deploy-time check.

### Checklist

### DIFF
--- a/tasks/dotfiles.py
+++ b/tasks/dotfiles.py
@@ -26,8 +26,8 @@ _dotfiles_dir = os.path.expanduser(host.data.dotfiles_dir)
 # Clone or update the dotfiles repository
 # ---------------------------------------------------------------------------
 # Mirrors the clone-or-pull pattern from bootstrap.sh: check for .git to
-# decide between clone and pull. --ff-only prevents merge commits if the
-# local branch has diverged unexpectedly.
+# decide between clone and pull. os.path.isdir() evaluates at control time
+# (Python parse), not deploy time — safe because Hanzo always targets @local.
 if os.path.isdir(os.path.join(_dotfiles_dir, ".git")):
     server.shell(
         name="Pull latest dotfiles changes",
@@ -88,6 +88,7 @@ else:
 # agents, settings.json). The repo uses an ignore-all .gitignore that
 # whitelists only config files, so untracked runtime files (memory/,
 # projects/, credentials) are preserved across all code paths.
+# Expanded on the control machine — safe because Hanzo always targets @local.
 _claude_config_dir = os.path.expanduser(host.data.claude_config_dir)
 
 if os.path.isdir(os.path.join(_claude_config_dir, ".git")):
--- a/tasks/dotfiles.py
+++ b/tasks/dotfiles.py
@@ -17,19 +17,7 @@ import shlex
 from pyinfra import host, logger
 from pyinfra.operations import server
 
-
-def _git_pull_ff_cmd(repo_dir: str) -> str:
-    """Build a git pull --ff-only command with a graceful fallback on divergence.
-
-    Uses --ff-only to avoid accidental merge commits. On failure (diverged
-    branch, network error, etc.) warns and continues rather than aborting the
-    entire provisioning run — a stale checkout is acceptable; a failed
-    provision is not.
-    """
-    return (
-        f"git -C {shlex.quote(repo_dir)} pull --ff-only"
-        " || echo 'WARNING: fast-forward failed; local branch may have diverged — skipping pull' >&2"
-    )
+from tasks.helpers import shell_git_pull
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


tasks/dotfiles.py
```diff
@@ -17,6 +17,21 @@
 from pyinfra import host, logger
 from pyinfra.operations import server
 
+
+def _git_pull_ff_cmd(repo_dir: str) -> str:
```
palazzem: I would reorganize the entire repo to have a `helpers` module so we avoid awkward `_functions`. Also call this just `shell_git_pull`.

tasks/packages.py
```diff
@@ -49,7 +51,8 @@
 server.shell(
     name="Install AUR packages via paru",
     commands=[
-        "paru -S --needed --noconfirm " + " ".join(host.data.aur_packages),
+        "paru -S --needed --noconfirm "
```
palazzem: do the same for `paru` as you did for `git_pull`? Follow my previous comment

tasks/tools.py
```diff
@@ -86,7 +87,7 @@
     server.shell(
         name=f"Install {_tool} via uv tool",
         commands=[
-            f'uv tool list 2>/dev/null | grep -q "^{_tool} " || uv tool install {_tool}',
+            f'uv tool list 2>/dev/null | grep -q "^{_tool} " || uv tool install {shlex.quote(_tool)}',
```
palazzem: same here, in `helpers`

tasks/tools.py
```diff
@@ -129,7 +131,7 @@
         name=f"Install {_pkg} via npm",
         commands=[
             f"{_FNM_ACTIVATE} && "
-            f"(npm list -g {_pkg} >/dev/null 2>&1 || npm install -g {_pkg})",
+            f"(npm list -g {shlex.quote(_pkg)} >/dev/null 2>&1 || npm install -g {shlex.quote(_pkg)})",
```
palazzem: Overall, I would ultrathink a better approach. Abstract all commands so that they are well defined in a helper file. This reduces duplication, and forces the same pattern across all tasks. This applies mostly to `server.shell` commands.

palazzem:
Addressed all review comments in commit 0fdbbc3. Created tasks/helpers.py with shell_git_pull, shell_paru_install, shell_uv_tool_install, shell_npm_install, and FNM_ACTIVATE. Refactored dotfiles.py, packages.py, and tools.py to use helpers. Lint and container build pass.

tasks/dotfiles.py
palazzem: Created tasks/helpers.py with public helpers. Moved _git_pull_ff_cmd there as shell_git_pull. Added tasks/__init__.py for standard Python imports.

tasks/packages.py
palazzem: Extracted shell_paru_install(packages) into tasks/helpers.py. Both packages.py and tools.py now import and use it. packages.py no longer needs import shlex directly.

tasks/tools.py
palazzem: Extracted shell_uv_tool_install(tool) into tasks/helpers.py. The loop in tools.py now calls the helper.

tasks/tools.py
palazzem: tasks/helpers.py now has all reusable server.shell builders: shell_git_pull, shell_paru_install, shell_uv_tool_install, shell_npm_install, plus FNM_ACTIVATE constant. One-off commands (pyenv, rustup, gcloud, hardware) stay inline since they have no reuse. All shlex.quote() is centralized in helpers. Lint and container build pass.

PR #239, opened 2026-05-13
fix(security): harden bootstrap config write and deploy key allowlist

palazzem:
### Related Issues

n/a

### Proposed Changes:

Two independent attack surfaces in the bootstrap/deploy pipeline are closed:

**`bin/bootstrap.sh` — config write hardening**

- Extracts a `write_config()` function, eliminating the TOCTOU race where the config directory was created before the file was written. The file is now created with `umask 077` in a subshell so it is never world-readable even briefly.
- Adds a `log_error()` helper that writes to stderr.
- Validates that `HANZO_FULLNAME` and `HANZO_EMAIL` do not contain newlines before writing them to the config file. A newline would allow injecting arbitrary `KEY=value` lines into the config.
- Double-quote escaping is consolidated into `write_config()` so both unattended and interactive paths go through identical sanitisation.

**`deploy.py` — config key allowlist**

- Introduces `_ALLOWED_CONFIG_KEYS`, an explicit set of the keys that `deploy.py` accepts from `~/.config/hanzo/config`. Any key not in the set is logged as a warning and silently dropped, preventing a malicious or corrupted config file from injecting arbitrary attributes into `host.data`.
- Replaces the vague "Phase 2 tasks" comment with an ASCII dependency graph that explains why task inclusion order matters.

### Testing:

- A new `RUN` step in `tests/Containerfile` writes a config containing an unknown key (`EVIL_KEY`) and asserts that `pyinfra` logs "Ignoring unknown config key: evil_key", verifying the allowlist is enforced end-to-end.
- Existing container smoke-test (`pyinfra -y @local deploy.py --dry`) continues to cover the happy path.

### Extra Notes (optional):

The `write_config()` extraction also removes duplicate escaping logic that previously lived separately in the unattended and interactive branches, reducing surface area for future drift.

### Checklist

### DIFF
--- a/tests/Containerfile
+++ b/tests/Containerfile
@@ -24,7 +24,8 @@ ENV SKIP_HARDWARE_CHECK=1
 RUN pyinfra -y @local deploy.py --dry
 
 # Exercises: dotfiles pull branch (existing .dotfiles), identity injection,
-# and claude-config pull branch (existing .claude repo)
+# and claude-config pull branch (existing .claude repo).
+# Also serves as the positive-path test for the config key allowlist.
 RUN mkdir -p /home/testuser/.dotfiles/.git && \
     mkdir -p /home/testuser/.claude/.git && \
     mkdir -p /home/testuser/.config/hanzo && \
--- a/tests/Containerfile
+++ b/tests/Containerfile
@@ -31,3 +31,8 @@ RUN mkdir -p /home/testuser/.dotfiles/.git && \
     printf 'HANZO_FULLNAME="Test User"\nHANZO_EMAIL="test@example.com"\n' \
       > /home/testuser/.config/hanzo/config && \
     pyinfra -y @local deploy.py --dry
+
+# Exercises: unknown config keys are rejected by the allowlist
+RUN printf 'HANZO_FULLNAME="Test User"\nHANZO_EMAIL="test@example.com"\nEVIL_KEY="payload"\n' \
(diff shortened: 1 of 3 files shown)


tests/Containerfile
```diff
@@ -24,10 +24,16 @@ ENV SKIP_HARDWARE_CHECK=1
...
-# and claude-config pull branch (existing .claude repo)
+# and claude-config pull branch (existing .claude repo).
+# Also serves as the positive-path test for the config key allowlist.
 RUN mkdir -p /home/testuser/.dotfiles/.git && \
     mkdir -p /home/testuser/.claude/.git && \
     mkdir -p /home/testuser/.config/hanzo && \
     printf 'HANZO_FULLNAME="Test User"\nHANZO_EMAIL="test@example.com"\n' \
       > /home/testuser/.config/hanzo/config && \
     pyinfra -y @local deploy.py --dry
+
```
palazzem: This is not a test. `Containerfile` must simply run what users run. Period, it's used to see if it works. If you need dedicated tests, write new Containerfiles.

palazzem: Removed both the test-oriented comment (line 28) and the EVIL_KEY assertion block (lines 36-39). The Containerfile now only runs what users run — two dry-run scenarios exercising real config paths. The allowlist in deploy.py still works; it just doesn't have an assertion in this Containerfile.

PR #253, opened 2026-05-16
feat(hardware): disable MT7925 Wi-Fi powersave via NM drop-in

palazzem:
### Related Issues

N/A — hardware quirk fix, no tracking issue.

### Proposed Changes

Disable NetworkManager Wi-Fi power-save (`wifi.powersave = 2`) for the MediaTek
MT7925 Wi-Fi chip in the GZ302. Implemented as a NetworkManager `conf.d/`
drop-in deployed by `roles/hardware` and gated by the existing GZ302 DMI +
container detection, so it only applies on real GZ302 hardware.

**Rationale**

The MT7925 Wi-Fi chip in the GZ302 has well-documented stability issues with
NetworkManager's default `wifi.powersave=3` (enabled): disconnects, throttled
upload (~200 Mbps vs 1.2 Gbps), slow speeds. Reported on Garuda, Fedora 42,
Arch, Framework 13, and Z13 2025 forums; a 2023 `mt76` patch attempted
driver-side default-off but NM still flips it on at the link layer.

This is hardware-driven (MT7925 quirk), not a generic preference, so applying
it conditionally on GZ302 detection is correct. The base CachyOS image ships
exactly one `NetworkManager/conf.d/` file (`dns.conf`); no `wifi.powersave`
override exists. Files in `/etc/NetworkManager/conf.d/` override `/usr/lib/`,
so our drop-in is not redundant with or shadowed by upstream.

Value `2` per NetworkManager docs: 0=default, 1=ignore, 2=disable, 3=enable.

**What changed**

- New template: `roles/hardware/templates/wifi-powersave.conf.j2`
- New variable: `gz302_wifi_powersave: 2` in `group_vars/all.yml`
- New task in `roles/hardware/tasks/gz302.yml`
- New handler `Reload NetworkManager` in `roles/hardware/handlers/main.yml`
  using `ansible.builtin.systemd_service` (`state: reloaded`)

### Testing

- [x] `docker build -f tests/Containerfile -t hanzo:test .` passes; `gz302.yml`
      is correctly skipped in the container guard path
- [ ] Post-merge on real GZ302 hardware: `cat /etc/NetworkManager/conf.d/wifi-powersave.conf`
      confirms the drop-in is in place
- [ ] On first deploy, reconnect Wi-Fi (or reboot) before running
      `nmcli -f 802-11-wireless.powersave c show <conn>` — `[connection]`
      defaults are applied at connection-activation time; `state: reloaded`
      (SIGHUP) does not retroactively apply `wifi.powersave` to already-active
      connections.

### Extra Notes (optional)

The handler uses `state: reloaded` (SIGHUP) which reloads NM config files
without dropping connections. The power-save setting takes effect on the next
connection activation, so a manual reconnect or reboot is needed on first
deploy to observe the change.

### Checklist

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -147,7 +147,13 @@ gz302_keyboard_scancode: "70072"
 gz302_keyboard_target_key: prog1
 
 # NetworkManager wifi.powersave for MT7925. NM's default (3 = enabled)
-# causes disconnects and throttled throughput on this chip.
+# causes disconnects and throttled throughput on this chip; 2 = disabled is
+# the cross-distro workaround (Arch, Fedora, Framework, Z13 communities).
+# Trade-off: ~1-2% extra system idle power; does NOT affect suspend (the
+# mt7925_pci_suspend driver callback is a separate code path from runtime
+# nl80211 PS). 1 = ignore is not a substitute because the mt76 driver
+# default is PS-on (the 2023 patch to flip it was rejected upstream).
+# Flip back to 3 if a future kernel/firmware combo fixes the regression.
 # Values: 0 = use default, 1 = ignore, 2 = disable, 3 = enable.
 gz302_wifi_powersave: 2
 
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -146,6 +146,10 @@ gz302_asus_lightbar_product: "18c6"
 gz302_keyboard_scancode: "70072"
 gz302_keyboard_target_key: prog1
 
+# NetworkManager wifi powersave for MT7925: 2 = disabled. NM default is 3
+# (enabled), which causes disconnects/slow throughput on this chip.
+gz302_wifi_powersave: 2
+
 # ---------------------------------------------------------------------------
 # Development tools (roles/devtools, roles/infra)
 # ---------------------------------------------------------------------------
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -146,8 +146,9 @@ gz302_asus_lightbar_product: "18c6"
 gz302_keyboard_scancode: "70072"
 gz302_keyboard_target_key: prog1
 
-# NetworkManager wifi powersave for MT7925: 2 = disabled. NM default is 3
-# (enabled), which causes disconnects/slow throughput on this chip.
+# NetworkManager wifi.powersave for MT7925. NM's default (3 = enabled)
+# causes disconnects and throttled throughput on this chip.
+# Values: 0 = use default, 1 = ignore, 2 = disable, 3 = enable.
 gz302_wifi_powersave: 2
 
 # ---------------------------------------------------------------------------
--- a/roles/hardware/handlers/main.yml
+++ b/roles/hardware/handlers/main.yml
@@ -19,3 +19,9 @@
     udevadm trigger
   changed_when: true
   become: true
+
+- name: Reload NetworkManager
+  ansible.builtin.systemd_service:
+    name: NetworkManager
+    state: reloaded
+  become: true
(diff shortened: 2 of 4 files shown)


group_vars/all.yml
```diff
@@ -146,6 +146,11 @@ gz302_asus_lightbar_product: "18c6"
 gz302_keyboard_scancode: "70072"
 gz302_keyboard_target_key: prog1
 
+# NetworkManager wifi.powersave for MT7925. NM's default (3 = enabled)
+# causes disconnects and throttled throughput on this chip.
+# Values: 0 = use default, 1 = ignore, 2 = disable, 3 = enable.
+gz302_wifi_powersave: 2
```
palazzem, line 158: Is this wrong though? Would it prevent sleeping / suspend issues or major battery consumption? Ensure you investigate it well, especially for GZ302 model (Flow Z13)

palazzem: Investigated. Three concerns, three answers:

**1. Suspend / sleep — non-issue.** `wifi.powersave` is a *runtime* nl80211 PS setting (PS-Poll / null-frame behavior). Suspend runs through the driver's PCI PM callback `mt7925_pci_suspend`, an entirely separate code path. Setting it to 2 cannot block suspend or break resume — confirmed by NM docs and the [mt7925 PCI suspend bugfix](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2141198) (commit `09e5c8afe92f`), which patches that callback independently of any `wifi.powersave` value. The two are orthogonal. Real GZ302 suspend bugs (mt7925_pci_suspend -110 timeout, Bluetooth-via-USB autosuspend on the same combo card, amdgpu PSR/Panel Replay) live in other code paths.

**2. Battery — small, quantified.** Generic 802.11 PSM literature puts PS savings at 10–20% of Wi-Fi-component draw ([Wi-Fi Alliance WMM-PS FAQ](https://www.wi-fi.org/knowledge-center/faq/how-does-wmm-power-save-work)). On a Wi-Fi 6E radio idling at 100–500 mW, PS-off costs roughly 100–300 mW extra. Against the Z13's ~12–18 W idle and 70 Wh battery, that's about **1–2% of system power** — minutes of runtime, not hours. The trade-off is the documented one across Arch, Fedora, Framework, and Z13 communities, who have all converged on disabling PS for this chip. I've also enriched the variable comment in `group_vars/all.yml` to capture this trade-off in-source so a future maintainer doesn't have to chase down the answer.

**3. Why `= 2` and not `= 1` (ignore).** I considered deferring to the driver default via `= 1`. It would be wrong: the mt76 driver default is PS **on**. Mario Limonciello's [Dec 2023 patch](https://patchwork.kernel.org/project/linux-wireless/patch/20231212090852.162787-1-mario.limonciello@amd.com/) ("wifi: mt76: mt7921/mt7925: Disable powersaving by default") was **rejected** by Felix Fietkau; no replacement merged. So `= 1` would leave PS enabled — exactly the broken state. `= 2` is the only correct choice today.

**Caveat worth flagging (out of scope for this PR):** part of the throughput collapse on related MediaTek chips traces to PCIe ASPM L1, not 802.11 PS — see [jetm.github.io on MT7927](https://jetm.github.io/blog/posts/mt7927-wifi-making-it-work/) and the Z13 community [mt7925-wifi-patches script](https://github.com/burakgon/mt7925-wifi-patches), which disables both PS *and* ASPM. The NM drop-in here fixes the PS half. If real-hardware testing post-merge shows residual throughput issues, the ASPM half can be a separate follow-up (likely `mt7925e.disable_aspm=1` module parameter via `/etc/modprobe.d/`).

**Escape hatch preserved:** `gz302_wifi_powersave` is exposed in `group_vars/all.yml`, so flipping it to 3 on a future kernel/firmware combo that fixes the regression is a one-line change, no role edit needed.

Issue #254, opened 2026-05-16
track: GZ302 hardware fixes to revisit (kernel & firmware triggers)


palazzem:
Tracking checklist for GZ302 hardware fix re-evaluations triggered by future kernel releases or firmware updates. Compiled after the 5-PR cleanup pass (PRs #249–#253) that aligned `roles/hardware/` with Linux 7.0 / CachyOS-gcc 7.0.8 baseline.

Each item names the trigger, the file/variable to change, and a pointer to the original analysis or PR discussion.

## Kernel-version-triggered reviews

- [ ] **Linux 7.1 lands in `linux-cachyos-gcc`** — review the suspend hook's ASUS HID unbind/rebind block in `roles/hardware/templates/gz302-suspend.sh.j2`.
  Denis Benato's hid-asus robustness series ([lore.kernel.org](https://lore.kernel.org/r/20260228191010.3830758-1-denis.benato@linux.dev/)) is queued for `for-7.1/asus` and targets the `asus_resume → asus_kbd_set_report → kmemdup → -ENOMEM` cascade we work around. Test 5+ resume cycles **without** the \`--- 2. Unbind ASUS HID devices ---\` (pre) and \`--- 4. Rebind ASUS HID devices ---\` (post) sections. If clean, delete those two sections. Leave NHI, xHCI, MMC steps alone — nobody upstream is working on those for Strix Halo.

- [ ] **Linux 7.2 lands in `linux-cachyos-gcc`** — review \`amdgpu.dcdebugmask=0x400\` (PR #251).
  The 7.2 AMDGPU Power Module rework ([Phoronix](https://www.phoronix.com/news/Linux-7.2-AMDGPU-Power-Mod)) realigns Replay/backlight/PSR with Windows — realistic milestone for Panel Replay being trustworthy on DCN 3.5 / Strix Halo. Test with \`gz302_oled_kernel_param: \"\"\` for several days under heavy scroll. If no flicker returns, delete the variable, the \`Ensure OLED kernel param in cmdline\` task in \`gz302.yml\`, and the \`Rebuild initramfs\` handler if no other consumer remains.
  *Note (2026-08-17): #273 accidentally reverted this value to \`0x600\` (bundled into an unrelated wifi-ordering fix); restored to \`0x400\` in #298. Trigger unchanged — still waiting on Linux 7.2.*

[...]

PR #256, opened 2026-05-17
fix(hardware): drop redundant RGB udev rule and harden suspend hook

palazzem:
## Summary

Two changes to the GZ302 hardware role, in this PR:

1. **Drop the GZ302 RGB udev rule entirely.** Both `0b05:1a30` (keyboard) and `0b05:18c6` (back-panel RGB / "lightbar") rules are removed, including the template file and the Ansible task that deployed it. A new `state: absent` task cleans up the previously deployed `/etc/udev/rules.d/99-gz302-rgb.rules` from hosts that already ran an earlier `hanzo`.
2. **Harden the suspend hook's `STATE_DIR` creation.** Replace `mkdir -p` with `install -d -m 0700 -o root -g root`.

### Why drop the RGB rule entirely (not just narrow it)

PR review on #256 challenged the premise of the rule. Investigation against the asusctl upstream architecture docs confirmed the rule grants permissions no userspace process actually needs:

- **`rog-control-center` (GUI)** communicates with the `asusd` daemon **exclusively over D-Bus** — it does not open USB or hidraw devices directly. `asusd` holds the device file handles.
- **`asusctl` (CLI)** also goes through D-Bus to `asusd` for hardware operations.
- **`asusd` runs as root** via systemd (`system_services: asusd` in the role's config), so it has full device access without any uaccess tag.
- **Upstream ships no `uaccess` tag** for `0b05:1a30` or `0b05:18c6`. PR #249's own description acknowledged this: *"CachyOS-Settings ships zero ASUS udev rules; asusctl's `asusd.rules` only fires the `asusd.service` on `asus-nb-wmi` DMI match and does not grant `0b05:18c6` ACLs."*

The rationale that introduced the lightbar rule in PR #249 — *"A userspace daemon driving the lightbar would need root because its USB device has no session ACL"* — was wrong about which process drives the lightbar. `asusd` drives the lightbar (and the keyboard RGB), as root, over D-Bus, with no userspace device handle ever open.

**Net effect:** all three keystroke / raw-USB vectors a seat-local process previously had on the keyboard (`/dev/hidraw*`, `/dev/input/event*`, and the bulk USB endpoint under `/dev/bus/usb/`) are now closed. The keystroke-capture residual that the earlier "narrow the rule" approach left open is gone.

### Why harden the suspend hook's STATE_DIR creation

`mkdir -p "$STATE_DIR"` is replaced with `install -d -m 0700 -o root -g root "$STATE_DIR"`. Concrete wins:

- **Atomic mode + ownership in a single syscall.** `mkdir -p` uses the process umask (typically `0022`, yielding `0755`), so there is a brief window where the dir is world-readable before a follow-up `chmod 0700` would land. `install -d -m 0700` sets the mode at creation — no race.
- **Idempotent drift auto-correction.** If a previous run (or an operator) left `$STATE_DIR` with drifted mode/ownership, every suspend cycle re-applies `0700 root:root`.
- **Catches the regular-file-at-path edge case.** If `$STATE_DIR` exists as a non-directory file, `install -d` fails with a clear "exists but is not a directory" message.

What `install -d` does **not** do, and why this is acceptable:

- It does **not** fail on a pre-existing dir with wrong owner/mode — it silently re-applies the requested `0700 root:root`.
- It does **not** refuse to follow a symlink — it follows and chmods the target. (`mkdir -p` is also unsafe here; neither tool defends against an attacker with `/run` write capability, which already requires root.)

The threat model of "root-capable attacker pre-plants in `/run`" is thin (an attacker who has root has already won), so out-of-scope here.

The hook intentionally lacks `set -euo pipefail` (documented at lines 13–16 of the script — the hook must be best-effort across every suspend cycle). The `install` call has no `|| true` guard, which means a non-zero exit (regular-file-at-path case) propagates as visible bash errors on the subsequent `: > "$STATE_DIR/..."` writes, surfacing in `journalctl -t gz302-suspend`.

### CI gates

- [x] `pre-commit run --all-files` passes locally and in CI.
- [x] `docker build -f tests/Containerfile -t hanzo:test .` succeeds. **Note (coverage caveat):** the GZ302 hardware tasks are skipped inside the container per the DMI + `virtualization_type` gate in `roles/hardware/tasks/main.yml`. The container build runs `ansible-playbook --check --diff` (the project's canonical test path per `CLAUDE.md`) plus `ansible-lint` static parsing — which catches malformed Jinja, malformed YAML, and module-arg validation. It does **not** render templates against real GZ302 facts or load the udev rule against a live udev daemon.

### Real-hardware validation (post-merge, on the actual GZ302)

The project owner's stated testing philosophy is: *drop the rule, verify, re-add if something breaks.* This PR drops the rule; the verification happens on real hardware.

- [ ] Re-run `hanzo` on the host. Confirm the new `Remove legacy RGB udev rule` task fires once and notifies `Reload udev rules`. On a subsequent run, the task should report `ok` (idempotent — file already absent).
- [ ] Confirm the rule is gone:
  ```
  ls /etc/udev/rules.d/99-gz302-rgb.rules
  # Expected: No such file or directory
  ```
- [ ] Confirm hidraw / input/event / bulk-USB nodes for `0b05:1a30` and `0b05:18c6` no longer carry uaccess:
  ```
  udevadm info /sys/class/hidraw/hidraw* | grep -E 'CURRENT_TAGS|TAGS'
  udevadm info /sys/class/input/event* | grep -B1 -A1 'ID_VENDOR_ID=0b05'
  ```
  Expected: no `:uaccess:` for ASUS-vendor devices.
- [ ] **Confirm `asusctl` CLI still works** without root:
  ```
  asusctl led-mode static -c ff0000   # or another effect command
  asusctl profile -P Quiet
  ```
- [ ] **Open `rog-control-center` and confirm RGB controls still function** for whatever lighting is present on the SKU (keyboard backlight, back-panel RGB if applicable).
- [ ] **If `openrgb` is installed**, confirm it can still drive ASUS devices via `asusd`.
- [ ] If anything breaks: capture which device / which userspace tool, and we re-add the minimum-necessary rule.
- [ ] Take a suspend/resume cycle: `systemctl suspend`, wake, confirm `journalctl -t gz302-suspend` shows both phases completing cleanly with no `install: cannot create directory` or related errors. Confirm `sudo stat -c '%a %U %G' /run/gz302-suspend` shows `700 root root` while the dir exists (between pre- and post-resume).

### DIFF
--- a/roles/hardware/tasks/gz302.yml
+++ b/roles/hardware/tasks/gz302.yml
@@ -1,8 +1,5 @@
 ---
 # ASUS ROG Flow Z13 (GZ302) hardware fixes.
-#
-# Fixes: GPU modprobe, OLED kernel param, initramfs,
-# suspend hook, keyboard hwdb remap, RGB udev rule.
 
 - name: Deploy GPU modprobe config
   ansible.builtin.template:
@@ -35,11 +32,10 @@
   become: true
   notify: Update hwdb
 
-- name: Deploy RGB udev rule
-  ansible.builtin.template:
-    src: 99-gz302-rgb.rules.j2
-    dest: /etc/udev/rules.d/99-gz302-rgb.rules
-    mode: "0644"
+- name: Remove legacy RGB udev rule
+  ansible.builtin.file:
+    path: /etc/udev/rules.d/99-gz302-rgb.rules
+    state: absent
   become: true
   notify: Reload udev rules
 
--- a/roles/hardware/tasks/gz302.yml
+++ b/roles/hardware/tasks/gz302.yml
@@ -32,13 +32,6 @@
   become: true
   notify: Update hwdb
 
-- name: Remove legacy RGB udev rule
-  ansible.builtin.file:
-    path: /etc/udev/rules.d/99-gz302-rgb.rules
-    state: absent
-  become: true
-  notify: Reload udev rules
-
 - name: Deploy NetworkManager wifi powersave drop-in
   ansible.builtin.template:
     src: wifi-powersave.conf.j2
--- a/roles/hardware/templates/99-gz302-rgb.rules.j2
+++ b/roles/hardware/templates/99-gz302-rgb.rules.j2
@@ -1,7 +1,28 @@
 # ASUS ROG Flow Z13 (GZ302) -- RGB unprivileged USB access
 # Managed by Hanzo. Do not edit manually.
 #
-# Grants unprivileged access to the ASUS keyboard and lightbar USB devices
-# so that rog-control-center can control RGB without running as root.
-SUBSYSTEMS=="usb", ATTRS{idVendor}=="{{ gz302_asus_usb_vendor }}", ATTRS{idProduct}=="{{ gz302_asus_keyboard_product }}", TAG+="uaccess"
-SUBSYSTEMS=="usb", ATTRS{idVendor}=="{{ gz302_asus_usb_vendor }}", ATTRS{idProduct}=="{{ gz302_asus_lightbar_product }}", TAG+="uaccess"
+# Grants unprivileged access to the ASUS keyboard and lightbar HID control
+# nodes so that rog-control-center can drive RGB without running as root.
+#
+# Why SUBSYSTEM (singular) and not SUBSYSTEMS (plural):
+# SUBSYSTEM== matches only the device's OWN subsystem; ATTRS{} still walks
+# parents to read USB vendor/product. SUBSYSTEMS=="usb" (plural) would walk
+# up to the USB parent and apply TAG+="uaccess" to every child node --
+# hidraw*, input/event*, and the bulk USB endpoint under /dev/bus/usb/ --
+# handing seat-local processes (Flatpak --device=all, browser native
+# messaging hosts, any desktop-user process) three separate keystroke /
+# raw-USB vectors on the keyboard (1a30).
+#
+# This rule eliminates TWO of those three vectors:
+#   * input/event* (the simpler, fully-decoded key-event source) -- CLOSED
+#   * bulk USB endpoint (arbitrary USB control transfers)        -- CLOSED
+#
+# Residual: the keyboard hidraw node ITSELF still carries uaccess, and HID
+# input reports can be decoded as keystrokes by a seat-local process. This
+# is accepted as a trade-off because rog-control-center / openrgb may need
+# direct keyboard hidraw access for capability detection, and we cannot
+# verify "rule is removable" without real-hardware testing. Tracked in
+# issue #254 -- if real-hardware validation shows the keyboard rule is
+# unused, dropping the 1a30 line is a one-line follow-up PR.
+SUBSYSTEM=="hidraw", ATTRS{idVendor}=="{{ gz302_asus_usb_vendor }}", ATTRS{idProduct}=="{{ gz302_asus_keyboard_product }}", TAG+="uaccess"
+SUBSYSTEM=="hidraw", ATTRS{idVendor}=="{{ gz302_asus_usb_vendor }}", ATTRS{idProduct}=="{{ gz302_asus_lightbar_product }}", TAG+="uaccess"
--- a/roles/hardware/templates/99-gz302-rgb.rules.j2
+++ b/roles/hardware/templates/99-gz302-rgb.rules.j2
@@ -1,9 +1,0 @@
-# ASUS ROG Flow Z13 (GZ302) -- RGB unprivileged USB access
-# Managed by Hanzo. Do not edit manually.
-#
-# Residual (see #254): the keyboard hidraw still carries uaccess, so HID
-# reports remain decodable as keystrokes by a seat-local process. The
-# 1a30 rule is kept until real-hardware testing confirms it is safe to
-# drop without breaking rog-control-center / openrgb capability detection.
-SUBSYSTEM=="hidraw", ATTRS{idVendor}=="{{ gz302_asus_usb_vendor }}", ATTRS{idProduct}=="{{ gz302_asus_keyboard_product }}", TAG+="uaccess"
-SUBSYSTEM=="hidraw", ATTRS{idVendor}=="{{ gz302_asus_usb_vendor }}", ATTRS{idProduct}=="{{ gz302_asus_lightbar_product }}", TAG+="uaccess"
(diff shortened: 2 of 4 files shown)


roles/hardware/templates/99-gz302-rgb.rules.j2
```diff
@@ -1,7 +1,9 @@
 # ASUS ROG Flow Z13 (GZ302) -- RGB unprivileged USB access
 # Managed by Hanzo. Do not edit manually.
 #
-# Grants unprivileged access to the ASUS keyboard and lightbar USB devices
-# so that rog-control-center can control RGB without running as root.
-SUBSYSTEMS=="usb", ATTRS{idVendor}=="{{ gz302_asus_usb_vendor }}", ATTRS{idProduct}=="{{ gz302_asus_keyboard_product }}", TAG+="uaccess"
-SUBSYSTEMS=="usb", ATTRS{idVendor}=="{{ gz302_asus_usb_vendor }}", ATTRS{idProduct}=="{{ gz302_asus_lightbar_product }}", TAG+="uaccess"
+# Residual (see #254): the keyboard hidraw still carries uaccess, so HID
```
palazzem: No PRs on comments. Also, drop the rule as the only way to test is "don't add, verify, add if it doesn't work".

palazzem: Also, what is "lightbar" in the GZ302? I don't have a lightbar, only the keyboard lighting. I also think rog-control-center is actually using asusd to control and pilot (it doesn't start without asusctl running), meaning I think neither of these are need. Investigate how rog-control-center and asusctl / asusd work. Then present me your discoveries to understand if we should remove it.

palazzem: Both points addressed in commit 6b50807 — pushed.

## On "drop the rule"

Done. The rule is removed in its entirety, not narrowed:

- `roles/hardware/templates/99-gz302-rgb.rules.j2` — **deleted**.
- `roles/hardware/tasks/gz302.yml` — the `Deploy RGB udev rule` task is replaced with a `Remove legacy RGB udev rule` task (`state: absent`, notifies `Reload udev rules`) so that hosts that already provisioned with prior `hanzo` runs get the deployed `/etc/udev/rules.d/99-gz302-rgb.rules` cleaned up on next run.

This closes all three keystroke / raw-USB vectors that the narrowing approach in this PR's earlier commits had only partially closed — `hidraw*`, `input/event*`, and the bulk USB endpoint under `/dev/bus/usb/` all lose `uaccess`. The PR title is updated to `fix(hardware): drop redundant RGB udev rule and harden suspend hook` and the PR body now reflects the simplified scope.

## On the rog-control-center / asusd architecture investigation

Verified your hypothesis is correct. Findings:

1. **`rog-control-center` does NOT open USB / hidraw directly.** Confirmed against the asusctl upstream architecture docs: *\"rog-control-center communicates exclusively through D-Bus and does not directly access USB or hidraw devices. The asusd daemon holds the device file handles, not rog-control-center.\"* No mention of \`uaccess\`, udev rules, or user-session device access requirements anywhere in the GUI source.
2. **`asusctl` (CLI) also goes through D-Bus.** Same source: *\"Both the CLI and GUI components communicate with the daemons via standardized D-Bus interfaces.\"* The CLI is not a special case.
3. **`asusd` is the device-handle holder.** It runs as root via systemd (already wired in this role through `asus_packages: asusctl` + `system_services: asusd`), so it has full device access without needing a uaccess tag.
4. **Upstream ships zero uaccess for these PIDs.** PR #249's own description acknowledged this: *\"CachyOS-Settings ships zero ASUS udev rules; asusctl's `asusd.rules` only fires the `asusd.service` on `asus-nb-wmi` DMI match and does not grant `0b05:18c6` ACLs.\"* That's the strongest data point — if upstream considered the ACL necessary, upstream would ship it.

PR #249's original rationale was *\"A userspace daemon driving the lightbar would need root because its USB device has no session ACL\"* — that premise was empirically wrong. Userspace daemons don't drive the lightbar at all; `asusd` drives it as root over its own D-Bus interface. Granting unprivileged session ACLs to those USB devices provided no functional capability that wasn't already available through `asusd`'s D-Bus surface — it only widened the keystroke-capture attack surface for seat-local processes.

## On \"lightbar\" terminology

The `0b05:18c6` device is the ROG Flow Z13's back-panel RGB (the RGB-lit logo / window on the back of the tablet). ASUS marketing describes the GZ302 as having *\"a CNC aluminum frame and features an RGB-lit back panel.\"* Some SKUs of the Z13 line ship the back-panel RGB and some don't — but this is moot for the udev rule decision: whether the device is physically present or not, `asusd` is the only thing that talks to it, and `asusd` runs as root.

The PID stays referenced in:

- `group_vars/all.yml:142` — `gz302_asus_lightbar_product: \"18c6\"` (kept; still used)
- `roles/hardware/templates/gz302-suspend.sh.j2:162` — the suspend hook cycles `authorized` on this USB device via `sysfs` writes as root to work around the ASUS HID `ENOMEM` re-probe bug on resume. That's a root sysfs write, not a userspace device open, so no uaccess is needed for it either.

If you confirm on real hardware that the GZ302 SKU in your possession definitely has no back-panel RGB at all, we could even drop the `gz302_asus_lightbar_product` variable and the lightbar case in the suspend hook's `reset_usb_device` loop — but that's a separate follow-up, not blocking on this PR.

## On \"No PRs on comments\"

Acknowledged as a project convention. The `(see #254)` reference in the previous version of `99-gz302-rgb.rules.j2` is no longer present (file is deleted). The remaining comment in the diff (`gz302-suspend.sh.j2:51-52`, the install-d caveat) contains no PR/issue references. Going forward I'll keep PR/issue cross-references in commit messages and PR descriptions only, never in source comments.

## Validation status

- `pre-commit run --all-files` — pass.
- `docker build -f tests/Containerfile -t hanzo:test .` — pass (`ok=36 changed=24 failed=0 skipped=24`).

Ready for re-review.

roles/hardware/tasks/gz302.yml
```diff
@@ -35,11 +32,10 @@
   become: true
   notify: Update hwdb
 
-- name: Deploy RGB udev rule
-  ansible.builtin.template:
-    src: 99-gz302-rgb.rules.j2
-    dest: /etc/udev/rules.d/99-gz302-rgb.rules
-    mode: "0644"
+- name: Remove legacy RGB udev rule
```
palazzem: Remove this, nothing has been deployed, so we want breaking changes to make things right.

palazzem: Done in commit `b7b57d7`. The `Remove legacy RGB udev rule` task is gone, and I also removed the `Reload udev rules` handler from `roles/hardware/handlers/main.yml` since the cleanup task was the only consumer — leaving it would be dead code. A future task that needs udev reload can re-add the handler when it has a real notifier.

Net change in this commit: **15 lines deleted, 0 added.**

Validation: `pre-commit` clean, container build clean.

PR #258, opened 2026-05-17
chore(packages,infra): supply-chain hygiene for 4 package entries

palazzem:
### Related Issues

N/A

### Proposed Changes

Consolidates four package-list corrections so the provisioner stops pulling third-party, unsigned, or nonexistent artifacts.

**`claude-desktop-bin` removed from `aur_packages`**
Anthropic does not ship a Linux desktop binary. The AUR entry is a third-party Electron wrapper; the actual workflow is covered by `@anthropic-ai/claude-code` (npm), which is already installed.

**`tfupdate-bin` dropped from `infra_aur_packages`**
Out-of-date since 2025-11-04, single maintainer, 2 votes. No replacement needed.

**`proton-vpn-gtk-app` moved from `security_aur_packages` → `security_packages` (pacman)**
The package ships Arch-signed in `extra/`. Routing it through AUR replaces a signed Arch binary with whatever the community packager produces — a strict trust downgrade. A new `security_packages` list handles packages from official repos that belong conceptually alongside security tooling.

**`terraform` AUR entry replaced with a GPG-verified direct-from-HashiCorp install (`roles/infra/tasks/terraform.yml`)**
`terraform-bin` (AUR) has no PGP verification in its PKGBUILD. HashiCorp publishes a GPG-signed `SHA256SUMS` file alongside every release. The new flow implements a four-gate trust chain:

1. Fetch HashiCorp's PGP key by URL; pin its CDN-served SHA256 as a defense-in-depth integrity check (cosmetic — CDN may rewrite trailing whitespace).
2. Import the key and assert the imported fingerprint matches `terraform_gpg_fingerprint`.
3. Assert that `gpg --export --export-options export-minimal` of the key hashes to `terraform_gpg_key_export_sha256` — the canonical, deterministic binary form of the key, CDN-immune. **This is the load-bearing gate**; it catches subkey tampering or wholesale key replacement even when the CDN-served file SHA happens to match.
4. Verify `SHA256SUMS` with `gpg --verify`, extract the trusted zip hash from the now-verified file, and pass it to `get_url` as the checksum gate for the zip download.

The trust-chain block runs with `check_mode: false` so the full chain is exercised end-to-end under `ansible-playbook --check` in CI; only the final `unarchive` into `/usr/local/bin` honors check mode.

**`unzip` added to `archive_packages`**
`ansible.builtin.unarchive` shells out to `unzip` for `.zip` archives. `unzip` is not in the CachyOS base image; without it, real-mode Terraform installs would fail at the unarchive step.

### Testing

- `pre-commit run --all-files` — ansible-lint, shellcheck, and generic hooks pass.
- `docker build -f tests/Containerfile -t hanzo:test .` — full `--check --diff` run inside the CachyOS container exercises the trust chain end-to-end (network fetch, key import, fingerprint + export-SHA assertions, `gpg --verify`, SHA extraction, zip download).

### Extra Notes

The `terraform_gpg_key_export_sha256` pin mirrors the pattern PR D adopts for the 1Password key: both use `gpg --export --export-options export-minimal | sha256sum` as a CDN-immune canonical gate, sharing the same reasoning.

A `terraform_cache_dir` variable (`~/.cache/hanzo`) is introduced to centralise the cache path — the directory is created by the playbook `pre_tasks` before this role runs.

### Checklist


group_vars/all.yml
```diff
@@ -70,10 +70,14 @@ virtualization_packages:
   - virt-viewer
   - virtiofsd
 
+# proton-vpn-gtk-app ships Arch-signed in extra/; AUR would replace it
```
palazzem: Unneeded comment.

group_vars/all.yml
```diff
@@ -167,9 +171,19 @@ uv_tools:
   - checkov
 
 infra_aur_packages:
-  - terraform
   - tflint-bin
-  - tfupdate-bin
+
+# Trust-chain rationale lives in roles/infra/tasks/terraform.yml.
```
palazzem: All of that is unneeded. There is https://archlinux.org/packages/extra/x86_64/terraform/ package you can install. It is flagged out-dated because 2 days ago a new version shipped. Remove all this bloated stuff as you can just install the extra package with pacman.

group_vars/all.yml
palazzem: Dropped in 5ec552b. Variable name + position next to `security_aur_packages` carries the same signal.

group_vars/all.yml
palazzem: You'\''re right — verified `terraform 1.15.2-1` is in extra/, signed by Justin Kromlinger. The plan only checked AUR ("never existed as an AUR package") and skipped the extra/ check, so it built a direct-from-HashiCorp trust chain instead of just using the Arch-signed pacman package. Net: same trust guarantees as every other extra/ install we already do, no ad-hoc machinery on top.

Fix in 5ec552b:
- Added `infra_packages: [terraform]` + a pacman install task (mirrors the `security_packages` convention).
- Deleted `roles/infra/tasks/terraform.yml` entirely.
- Dropped every `terraform_*` variable and the `include_tasks` line.
- Dropped `unzip` from `archive_packages` — no other unarchive consumer in the playbook.

Container test: `ok=38 changed=26 failed=0 skipped=24`.

group_vars/all.yml
```diff
@@ -51,7 +51,6 @@ desktop_packages:
   - spotify-launcher
 
 aur_packages:
-  - claude-desktop-bin
```
palazzem, line 54: You can use `claude-desktop-appimage` as this maintainer is trusted: https://github.com/aaddrick/claude-desktop-debian

Verify the AUR package is actually signed and maintained. Do a proper security review to ensure we install trusted software.

group_vars/all.yml
```diff
@@ -166,10 +167,11 @@ uv_tools:
   - pre-commit
   - checkov
 
-infra_aur_packages:
+infra_packages:
   - terraform
+
+infra_aur_packages:
   - tflint-bin
```
palazzem: use tflint on extra instead of AUR package

group_vars/all.yml
palazzem: Added in 19a2bd7. Trust review per your request:

- **Maintainership**: AUR page lists **Mathys Lopinto** as packager *and* **Aaddrick Williams** (the upstream author you flagged as trusted) as co-maintainer — so the upstream-project author is on the AUR maintainer list, which is a meaningfully stronger trust chain than `claude-desktop-bin` had (`patrickjaja` was a third-party repackager unrelated to upstream).
- **Integrity gate**: PKGBUILD pins `sha256sums` on the AppImage downloaded from `github.com/aaddrick/claude-desktop-debian/releases/...`. That's the appropriate trust mechanism for an AppImage from a tagged GitHub release.
- **No `validpgpkeys`** — but PGP isn't part of the GitHub Releases trust model here; the sha256sum + the upstream-author co-maintainership cover what PGP would.
- **Lower vote count** than the removed package (4 vs 14, popularity 0.82 vs 4.91) — flagging it as the one weaker signal. Net: trust chain is stronger but community-validation signal is thinner. Happy to revert if the maintainership distinction concerns you on closer look.

group_vars/all.yml
palazzem: Done in 19a2bd7. Verified extra/tflint = 0.61.0-1, signed by Arch maintainer Andreas Schleifer (flagged out-of-date 2026-05-04 — same catch-up lag as terraform was). Same fix shape as round 1:

- Moved `tflint` to `infra_packages` (alongside `terraform`).
- Dropped `tflint-bin` from `infra_aur_packages`. That list is now empty, so the variable and its `kewlfft.aur.aur` install task in `roles/infra/tasks/main.yml` are removed too — no orphaned no-op.

Container test: `ok=37 changed=25 failed=0 skipped=24`.

PR #261, opened 2026-05-17
refactor(roles): extract foundation role for cache, locale, baseline CLI

palazzem:
### Related Issues

No GitHub issue — internal task tracking only.

### Proposed Changes

**PR 3 of 11** in a serialized refactor of the hanzo Ansible role structure.

Extracts substrate state that every other role implicitly depends on into a dedicated `roles/foundation/` role. Foundation is wired as the **first** role in `playbook.yml` with `tags: [foundation, always]`, so it runs before anything else on every invocation — including selective `--tags <role>` runs.

**What moves into `roles/foundation/`:**

- `Ensure Hanzo cache directory exists` — migrated from `playbook.yml` `pre_tasks`. The task retains `check_mode: false` (so `--check` runs still create the dir) and `become: false` (user-space path).
- `Set system locale` — migrated from `roles/system/tasks/main.yml`. Variable renamed `system_locale` → `foundation_locale` per the role-name prefix rule (ansible-lint `var-naming[no-role-prefix]`).
- `Install baseline development packages` — migrated from `roles/packages/tasks/main.yml`. Variable renamed `base_packages` → `foundation_packages`. Foundation now owns the single `update_cache: true` pacman DB refresh for the whole provisioning run; subsequent `community.general.pacman` calls in `roles/packages/` rely on the fresh cache without re-doing the work.

**`uv` is intentionally absent from `foundation_packages`.**

`uv` was `base_packages[3]` in `group_vars/all.yml`. It is installed by `bin/bootstrap.sh` (line 51, real users) and `tests/Containerfile` (line 21, CI) via the Astral installer at `~/.local/bin/uv` **before** `ansible-playbook` runs. That install path means `roles/devtools`'s `uv tool install …` chain keeps working without pacman touching `uv`. A later PR will formalize `uv` install under a dedicated role; not in scope here.

**`always` tag rationale.**

`roles/trust/` (GPG cache paths), `roles/languages/` (pyenv installer), and `roles/infra/` (gcloud installer) all reference `~/.cache/hanzo`. Without `always`, a selective `--tags trust` or `--tags languages` run would fail with a missing parent directory. Tagging `[foundation, always]` guarantees the cache dir exists on every invocation.

**Stale CLAUDE.md edge bullet deleted.**

[...]

PR #264, opened 2026-05-17
refactor(languages): make languages role self-contained

palazzem:
### Related Issues

No tracked issue — internal roles refactor (PR 5 of 11).

### Proposed Changes

The `roles/languages/` role previously relied on `roles/packages/` to pre-install `rustup`, `go`, and `fnm-bin` via pacman/paru, and on `group_vars/all.yml` for the `gopath` variable. This split ownership made it hard to reason about language provisioning in isolation.

This PR makes the role fully self-contained:

- Adds `roles/languages/vars/main.yml` with per-language package lists (`languages_python_packages`, `languages_rust_packages`, `languages_go_packages`, `languages_node_packages`) and `languages_gopath`. All names are prefixed per the `var-naming[no-role-prefix]` rule.
- Splits the single flat `tasks/main.yml` into four focused task files: `python.yml`, `rust.yml`, `node.yml`, `go.yml`. Each file owns the full lifecycle for its language (package install + runtime configuration).
- Rewrites `tasks/main.yml` as an orchestrator of four `include_tasks` calls — no task logic remains in it.
- Drops `runtime_packages`, `fnm-bin` from `aur_packages`, and `gopath` from `group_vars/all.yml`.
- Removes the "Install language runtimes" task from `roles/packages/tasks/main.yml`.

`uv` is restored here (as `languages_python_packages`) after being intentionally dropped from `foundation_packages` in PR #261 as a hand-off to this PR.

### Testing

- Lint: `pre-commit run --all-files`
- Container check: `docker build -f tests/Containerfile -t hanzo:test .`
- Targeted check: `docker build --build-arg ANSIBLE_ARGS="--tags languages --check --diff" -f tests/Containerfile -t hanzo:test .`

### Extra Notes (optional)

The `when: not ansible_check_mode` guards on `rustup`, `fnm`, and `pyenv` tasks are preserved as-is — they rely on binaries installed earlier in the same run and check mode never actually installs them.

### Checklist

### DIFF
--- a/roles/languages/vars/main.yml
+++ b/roles/languages/vars/main.yml
@@ -11,4 +11,4 @@ languages_go_packages:
 languages_node_packages:
   - fnm-bin
 
-languages_gopath: ~/programs/go
+languages_gopath: ~/.golang
--- a/roles/languages/vars/main.yml
+++ b/roles/languages/vars/main.yml
@@ -1,0 +1,14 @@
+---
+languages_python_packages:
+  - uv
+
+languages_rust_packages:
+  - rustup
+
+languages_go_packages:
+  - go
+
+languages_node_packages:
+  - fnm-bin
+
+languages_gopath: ~/programs/go
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -14,10 +14,6 @@ asus_packages:
   - asusctl
   - rog-control-center
 
-runtime_packages:
-  - rustup
-  - go
-
 gaming_packages:
   - cachyos-gaming-applications
   - ffmpeg
@@ -30,7 +26,6 @@ desktop_packages:
   - spotify-launcher
 
 aur_packages:
-  - fnm-bin
   - google-chrome
   - wifiman-desktop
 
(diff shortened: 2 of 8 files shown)


roles/languages/vars/main.yml
```diff
@@ -0,0 +1,14 @@
...
+languages_rust_packages:
+  - rustup
+
+languages_go_packages:
+  - go
+
+languages_node_packages:
+  - fnm-bin
+
+languages_gopath: ~/programs/go
```
palazzem: just call it `~/.golang`.

palazzem: Done. Renamed `languages_gopath` to `~/.golang` in `roles/languages/vars/main.yml`. Verified no other references in the repo (grepped across roles, group_vars, docs, shell/fish configs). Selective `--tags languages --check --diff` confirms the GOPATH dirs resolve correctly to `/home/testuser/.golang` and `/home/testuser/.golang/bin`. Pushed as 754bec9.

Issue #11, opened 2013-07-09
Add postgresql devel package


palazzem:
Package: `libpq-dev`

PR #269, opened 2026-05-17
refactor(roles): self-contain virtualization, delete system role

palazzem:
### Related Issues

Internal task #11 — final step of the 11-PR role refactor series (PRs 1–10 already merged on main). No GitHub issue.

### Proposed Changes:

This is the concluding PR of the role-refactor series. It removes the `system` role entirely and makes `virtualization` self-contained.

**`roles/system` deleted** — the role existed solely to add the user to `kvm`/`libvirt` groups and manage `system_services` (an empty list). Both responsibilities are now handled where the groups originate.

**`roles/virtualization` absorbs group management** — two tasks added:
- `Ensure virtualization groups exist` (uses `ansible.builtin.group`, runs outside check mode so groups are present before the user-append step)
- `Add user to virtualization supplementary groups` (mirrors the old `system` task, scoped to virtualization)

**`roles/virtualization/vars/main.yml` introduced** — `virtualization_packages` and the new `virtualization_user_groups` live here, following the role-scoped data convention established across the series.

**`group_vars/all.yml` collapsed** — only the `gz302_*` hardware identifiers remain (the only variables that cannot carry a role prefix under Rule 9 without redundancy).

**`playbook.yml` updated** — `system` role entry removed. Playbook now declares 11 roles in this order: foundation, desktop, virtualization, containers, trust, languages, tools, ai, security, hardware, home.

**`CLAUDE.md` Rule 1 updated** — reflects the authoritative convention: role-scoped data in `roles/<role>/vars/main.yml`; `group_vars/all.yml` reserved for `gz302_*` hardware identifiers only.

### Testing:

- [x] `docker build --build-arg ANSIBLE_ARGS="--list-tags" -f tests/Containerfile -t hanzo:test .` — shows 11 role tags + `always`; no `system` tag present
- [x] `docker build --build-arg ANSIBLE_ARGS="--tags virtualization --check --diff" -f tests/Containerfile -t hanzo:test .` — group-ensure executing, user-append skipping (check mode), libvirt block skipping (container, no /dev/kvm)
- [x] `docker build -f tests/Containerfile -t hanzo:test .` — full real provisioning run passes (ok=50, changed=26)

### Extra Notes (optional):

The cross-role dependency comment that previously appeared in `roles/virtualization/tasks/main.yml` (stating that `roles/system` must run after virtualization to add group membership) is removed — the dependency no longer exists.

### Checklist


group_vars/all.yml
```diff
@@ -1,33 +1,4 @@
...
-  - kvm
-  - libvirt
-
-system_services: []
-
-# ---------------------------------------------------------------------------
-# GZ302 (ASUS ROG Flow Z13) hardware configuration (roles/hardware)
-# ---------------------------------------------------------------------------
-
 # Disable Panel Replay (DC_DISABLE_REPLAY=0x400) on the Strix Halo OLED panel
```
palazzem: Why are we keeping these? Can't we move these in `hardware` as we did for everything else? Clean it up and to the final refactoring, probably this all.yml would not be needed at all.

CLAUDE.md
```diff
@@ -4,7 +4,7 @@ CachyOS system provisioner powered by Ansible.
 
 ## Rules
 
-1. Data in `group_vars/all.yml`, logic in `roles/`. Never hardcode package names, paths, or config values in task files. Add them to `group_vars/all.yml` and access via the plain variable name.
+1. Role-scoped data in `roles/<role>/vars/main.yml`; logic in `roles/<role>/tasks/`. `group_vars/all.yml` is reserved for variables that would gain a redundant prefix under Rule 9 (currently the `gz302_*` hardware identifiers consumed by `roles/hardware`). Never hardcode package names, paths, or config values in task files — declare them as variables and reference by name.
```
palazzem: I think this rule has no sense. Also you are mentioning all.yml that you were supposed to remove entirely after this last refactoring step.

CLAUDE.md
```diff
@@ -4,7 +4,7 @@ CachyOS system provisioner powered by Ansible.
...
 2. Prefer Ansible modules (`community.general.pacman`, `ansible.builtin.template`, `ansible.builtin.systemd_service`, `ansible.builtin.user`, `ansible.builtin.lineinfile`) over `ansible.builtin.shell` / `ansible.builtin.command`. Modules are idempotent for free.
```
palazzem: Remove this rule, makes no sense.

CLAUDE.md
palazzem: seems redundant with rule (3)

group_vars/all.yml
palazzem: Done. Moved `gz302_*` into `roles/hardware/vars/main.yml` as `hardware_gz302_*` (Rule 9 prefix applied), renamed all 11 references across `roles/hardware/tasks/gz302.yml` and the four templates, and deleted `group_vars/all.yml` entirely. The directory is gone. Pushed in 9848fb4.

CLAUDE.md
palazzem: Removed. With `group_vars/all.yml` deleted, the rule's exception clause no longer has a referent and its other half just describes standard Ansible role layout. Pushed in 9848fb4.

CLAUDE.md
palazzem: Removed. Rule 2 (prefer modules) and Rule 3 (shell/command is the escape hatch) state the same policy from opposite directions — kept Rule 3 and dropped Rule 2. Pushed in 9848fb4.

CLAUDE.md
palazzem: Removed. ansible-lint enforces both elements of Rule 7 universally (`name[missing]` and `name[template]`); restating linter-enforced rules in the docs is decoration. Rule 3 covers the shell/command `name:` requirement. Pushed in 9848fb4.

PR #273, opened 2026-05-17
fix(hardware): patch wifi first to prevent TCP stalls

palazzem:
### Related Issues

N/A

### Proposed Changes:

Three coupled bugs in the GZ302 hardware role, fixed together because they share the same handler refactor:

**1. OLED kernel param written to wrong file**

`/etc/kernel/cmdline` (UKI convention) does not exist on a default CachyOS install. The cmdline lives in `/etc/default/limine` and is consumed by `limine-mkinitcpio`. The `lineinfile` task is retargeted at the correct path with the proper `KERNEL_CMDLINE[default]+=` syntax.

**2. Wifi-powersave fix applied too late**

The MT7925 wifi-powersave drop-in was the last task in `gz302.yml`, meaning paru/curl downloads in later roles (desktop, ai, security) were stalling on the powersave bug before the fix took effect. The task is now first in the role, followed immediately by `meta: flush_handlers`, so the fix is active before any download runs.

**3. `dcdebugmask` value too low (`0x400` → `0x600`)**

`0x400` disables Panel Replay but leaves PSR-SU active, which causes scrolling artifacts on the Strix Halo OLED panel. `0x600` disables both (Panel Replay + PSR-SU), per upstream guidance.

**Handler refactor (necessary for the above):**

- Old: `Rebuild initramfs` (mkinitcpio -P) + `Reload NetworkManager` (systemctl reload)
- New: `Restart network stack and cycle wifi` (NM restart + radio off/on + 5s reconnect wait) and `Regenerate initramfs and Limine entries` (runs `limine-mkinitcpio`, the CachyOS canonical command that does initramfs rebuild + Limine entries update in a single step)

The now-stale `meta: flush_handlers` after the hardware role in `playbook.yml` is removed; the wifi flush moved inside the role, and the boot-regen handler fires once at end-of-play via Ansible's handler dedup.

### Testing:

```
docker build --build-arg ANSIBLE_ARGS="--tags hardware --check --diff" \
  -f tests/Containerfile -t hanzo:test .
```

Verify in the diff output:
- `/etc/NetworkManager/conf.d/wifi-powersave.conf` is written first, handler flushed before ASUS packages install
- `/etc/default/limine` receives the `amdgpu.dcdebugmask=0x600` line (not `/etc/kernel/cmdline`)
- `limine-mkinitcpio` is queued (not `mkinitcpio -P`)

### Extra Notes (optional):

The three changes are tightly coupled through the handler rename; splitting them into separate commits would leave the role in a broken state between commits.

### Checklist

- [ ] Tests are defining the correct and expected behavior

### DIFF
--- a/playbook.yml
+++ b/playbook.yml
@@ -59,10 +59,6 @@
           become: true
           tags: [always]
 
-        # Hardware runs first so the MT7925 wifi-powersave fix is
-        # applied before any other role starts downloading. The wifi
-        # handler is flushed inside the role; the boot-regen handler
-        # fires once at end-of-play.
         - name: Run hardware role
           ansible.builtin.import_role:
             name: hardware
--- a/playbook.yml
+++ b/playbook.yml
@@ -59,24 +59,19 @@
           become: true
           tags: [always]
 
-        - name: Run foundation role
-          ansible.builtin.import_role:
-            name: foundation
-          tags: [foundation, always]
-
-        # Hardware runs early so the MT7925 wifi-powersave drop-in is in
-        # place before the AUR-heavy roles below start downloading.
-        # NM's default powersave (3) silently stalls long-running curl
-        # transfers on this chip — Firefox masks it, but paru/makepkg do
-        # not. flush_handlers forces "Reload NetworkManager" to fire now
-        # rather than at end-of-play.
+        # Hardware runs first so the MT7925 wifi-powersave fix is
+        # applied before any other role starts downloading. The wifi
+        # handler is flushed inside the role; the boot-regen handler
+        # fires once at end-of-play.
         - name: Run hardware role
           ansible.builtin.import_role:
             name: hardware
           tags: [hardware]
 
-        - name: Flush handlers so hardware fixes take effect immediately
-          ansible.builtin.meta: flush_handlers
+        - name: Run foundation role
+          ansible.builtin.import_role:
+            name: foundation
+          tags: [foundation, always]
 
         - name: Run desktop role
           ansible.builtin.import_role:
--- a/roles/hardware/handlers/main.yml
+++ b/roles/hardware/handlers/main.yml
@@ -1,5 +1,5 @@
 ---
-- name: Restart network stack and cycle wifi
+- name: Restart network stack
   ansible.builtin.shell: |
     set -euo pipefail
     systemctl restart NetworkManager
@@ -9,7 +9,7 @@
   changed_when: true
   become: true
 
-- name: Regenerate initramfs and Limine entries
+- name: Regenerate initramfs
   ansible.builtin.command: limine-mkinitcpio
   changed_when: true
   become: true
--- a/roles/hardware/handlers/main.yml
+++ b/roles/hardware/handlers/main.yml
@@ -1,11 +1,15 @@
 ---
-- name: Rebuild initramfs
-  ansible.builtin.command: mkinitcpio -P
+- name: Restart network stack and cycle wifi
+  ansible.builtin.shell: |
+    set -euo pipefail
+    systemctl restart NetworkManager
+    nmcli radio wifi off
+    nmcli radio wifi on
+    sleep 5
   changed_when: true
   become: true
 
-- name: Reload NetworkManager
-  ansible.builtin.systemd_service:
-    name: NetworkManager
-    state: reloaded
+- name: Regenerate initramfs and Limine entries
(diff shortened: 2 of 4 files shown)


roles/hardware/handlers/main.yml
```diff
@@ -1,11 +1,15 @@
 ---
-- name: Rebuild initramfs
-  ansible.builtin.command: mkinitcpio -P
+- name: Restart network stack and cycle wifi
```
palazzem: Just "Restart network stack"

roles/hardware/handlers/main.yml
```diff
@@ -1,11 +1,15 @@
...
+    nmcli radio wifi on
+    sleep 5
   changed_when: true
   become: true
 
-- name: Reload NetworkManager
-  ansible.builtin.systemd_service:
-    name: NetworkManager
-    state: reloaded
+- name: Regenerate initramfs and Limine entries
```
palazzem: Just "Regenerate initramfs"

roles/hardware/tasks/gz302.yml
```diff
@@ -30,27 +44,19 @@
...
+  notify: Regenerate initramfs and Limine entries
 
-- name: Ensure OLED kernel param in cmdline
+- name: Ensure OLED kernel param in Limine cmdline
   ansible.builtin.lineinfile:
-    path: /etc/kernel/cmdline
+    path: /etc/default/limine
     regexp: 'amdgpu\.dcdebugmask='
-    line: "{{ hardware_gz302_oled_kernel_param }}"
+    line: 'KERNEL_CMDLINE[default]+=" {{ hardware_gz302_oled_kernel_param }}"'
```
palazzem, line 53: KERNEL_CMDLINE has already other parameters set. Would this actually break/remove parameters we have by default?

playbook.yml
```diff
@@ -59,24 +59,19 @@
...
-            name: foundation
-          tags: [foundation, always]
-
-        # Hardware runs early so the MT7925 wifi-powersave drop-in is in
-        # place before the AUR-heavy roles below start downloading.
-        # NM's default powersave (3) silently stalls long-running curl
-        # transfers on this chip — Firefox masks it, but paru/makepkg do
-        # not. flush_handlers forces "Reload NetworkManager" to fire now
-        # rather than at end-of-play.
+        # Hardware runs first so the MT7925 wifi-powersave fix is
```
palazzem: Remove this ocmment, is useless

roles/hardware/handlers/main.yml
palazzem: Renamed to `Restart network stack`. Task notify updated to match.

roles/hardware/handlers/main.yml
palazzem: Renamed to `Regenerate initramfs`. Both task notify references updated to match.

roles/hardware/tasks/gz302.yml
palazzem: Valid concern — confirmed. With the original loose regex `amdgpu\.dcdebugmask=`, lineinfile would have replaced the entire matched line. A line like `KERNEL_CMDLINE[default]+=" amdgpu.dcdebugmask=0x400 amd_pstate=guided"` would lose `amd_pstate=guided` on every run.

Tightened the regex to `^KERNEL_CMDLINE\[default\]\+="\s*amdgpu\.dcdebugmask=\S+\s*"\s*$` — it only matches a **dedicated, single-param Hanzo-shaped line**. Behavior:

- **First run, dcdebugmask absent anywhere** → regex no match → lineinfile appends a new dedicated `KERNEL_CMDLINE[default]+=" amdgpu.dcdebugmask=0x600"` line. Existing KERNEL_CMDLINE lines untouched.
- **Re-run with the Hanzo-managed line present** → regex matches → line replaced in place with the new value. Idempotent.
- **dcdebugmask exists inline mixed with other params** → regex doesn't match → lineinfile appends our own dedicated line. Original mixed line stays; bash chains the `+=` so kernel cmdline parser sees the last `amdgpu.dcdebugmask=` (ours) and uses that value. **No data loss.**

Trade-off: in the mixed-inline case the old value lingers in the file until the user cleans it up, but it never wins at boot. Avoiding the destructive replace was the priority.

playbook.yml
palazzem: Removed.

PR #275, opened 2026-05-22
feat(hardware): enable suspend-then-hibernate

palazzem:
Provisions the chain that converts lid-close into a deferred hibernate:

- /swap BTRFS subvolume + swapfile via `btrfs mkswapfile` (enforces NOCOW and no-compression, both required for hibernation I/O), sized from ansible_memtotal_mb + 4 GB so it adapts across RAM-variant SKUs.
- fstab entry and conditional swapon (no-op if already active).
- /etc/systemd/sleep.conf.d/ drop-in with HibernateDelaySec=1800.
- /etc/systemd/logind.conf.d/ drop-in with HandleLidSwitch= suspend-then-hibernate. Inert under KDE (PowerDevil holds a block-mode handle-lid-switch inhibitor) but applies for TTY, recovery, and non-KDE sessions.
- ~/.config/powerdevilrc SleepMode=3 under AC and Battery - the setting that actually drives behavior whenever Plasma is running.

Resume relies on systemd-hibernate-resume reading the HibernateLocation EFI variable, so no resume=/resume_offset= kernel cmdline params are needed. logind reload uses SIGHUP rather than service restart so active sessions survive.

### Checklist

- [ ] Related issue and proposed changes are filled
- [ ] Tests are defining the correct and expected behavior
- [ ] Commits follow [Conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/)

### DIFF
--- a/roles/hardware/tasks/gz302.yml
+++ b/roles/hardware/tasks/gz302.yml
@@ -95,18 +95,11 @@
     state: present
   become: true
 
-- name: Check whether swap is already active
-  ansible.builtin.command: swapon --show=NAME --noheadings
-  register: hardware_gz302_active_swaps
+# swapon --all activates everything in /etc/fstab and silently skips
+# devices already in use, so the task is idempotent without a pre-check.
+- name: Activate swap from fstab
+  ansible.builtin.command: swapon --all
   changed_when: false
-  check_mode: false
-  become: true
-
-- name: Activate swap if not already active
-  ansible.builtin.command:
-    cmd: swapon /swap/swapfile
-  when: "'/swap/swapfile' not in hardware_gz302_active_swaps.stdout"
-  changed_when: true
   become: true
 
 # systemd-suspend-then-hibernate.service consults HibernateDelaySec at
@@ -143,19 +136,16 @@
 
 # KDE Plasma 6: PowerDevil writes lid-action behavior here. SleepMode=3
 # = "Suspend then Hibernate" — set under both AC and Battery so behavior
-# is consistent regardless of power source. Section name is the literal
-# KDE-style nested form `[AC][SuspendAndShutdown]`; ini_file emits this
-# correctly when passed as a single section string with embedded `][`.
-- name: Configure KDE PowerDevil sleep mode (AC and Battery)
-  community.general.ini_file:
-    path: "{{ ansible_facts.env.HOME }}/.config/powerdevilrc"
-    section: "{{ item }}][SuspendAndShutdown"
-    option: SleepMode
-    value: "{{ hardware_gz302_kde_sleep_mode }}"
+# is consistent regardless of power source.
+- name: Configure KDE PowerDevil sleep mode
+  ansible.builtin.copy:
+    dest: "{{ ansible_facts.env.HOME }}/.config/powerdevilrc"
+    content: |
+      # Managed by Hanzo. Do not edit manually.
+      [AC][SuspendAndShutdown]
+      SleepMode=3
+
+      [Battery][SuspendAndShutdown]
+      SleepMode=3
     mode: "0644"
-    no_extra_spaces: true
-    create: true
-  loop:
-    - AC
-    - Battery
   become: false
--- a/roles/hardware/tasks/gz302.yml
+++ b/roles/hardware/tasks/gz302.yml
@@ -70,3 +70,92 @@
     dest: /usr/lib/systemd/system-sleep/gz302-suspend.sh
     mode: "0755"
   become: true
+
+# Hibernation backing store. Single contiguous swap file on a dedicated
+# NOCOW subvolume (CachyOS recommended layout). The kernel reads the
+# image back via a direct device offset systemd records in the
+# HibernateLocation EFI variable, so no resume= cmdline param is needed.
+- name: Ensure /swap btrfs subvolume exists
+  ansible.builtin.command:
+    cmd: btrfs subvolume create /swap
+    creates: /swap
+  become: true
+
+- name: Ensure /swap/swapfile exists
+  ansible.builtin.command:
+    cmd: btrfs filesystem mkswapfile --size {{ hardware_gz302_swap_size }} --uuid clear /swap/swapfile
+    creates: /swap/swapfile
+  become: true
+
+- name: Register swap in /etc/fstab
+  ansible.builtin.lineinfile:
+    path: /etc/fstab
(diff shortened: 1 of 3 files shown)


roles/hardware/tasks/gz302.yml
```diff
@@ -70,3 +70,92 @@
...
+
+- name: Register swap in /etc/fstab
+  ansible.builtin.lineinfile:
+    path: /etc/fstab
+    line: "/swap/swapfile none swap defaults 0 0"
+    regexp: '^/swap/swapfile\s'
+    state: present
+  become: true
+
+- name: Check whether swap is already active
```
palazzem: Uneeded, swapon must run and will be the command to skip if the swap is already enabled.

roles/hardware/tasks/gz302.yml
```diff
@@ -70,3 +70,92 @@
...
+  ansible.builtin.command: swapon --show=NAME --noheadings
+  register: hardware_gz302_active_swaps
+  changed_when: false
+  check_mode: false
+  become: true
+
+- name: Activate swap if not already active
+  ansible.builtin.command:
+    cmd: swapon /swap/swapfile
+  when: "'/swap/swapfile' not in hardware_gz302_active_swaps.stdout"
```
palazzem: if swap file already exists, this is realistically skipped.

roles/hardware/tasks/gz302.yml
```diff
@@ -70,3 +70,92 @@
...
+    mode: "0644"
+  become: true
+  notify: Reload logind
+
+# KDE Plasma 6: PowerDevil writes lid-action behavior here. SleepMode=3
+# = "Suspend then Hibernate" — set under both AC and Battery so behavior
+# is consistent regardless of power source. Section name is the literal
+# KDE-style nested form `[AC][SuspendAndShutdown]`; ini_file emits this
+# correctly when passed as a single section string with embedded `][`.
+- name: Configure KDE PowerDevil sleep mode (AC and Battery)
```
palazzem: Use the same pattern above with copy. No need to use a special ini_file task

roles/hardware/vars/main.yml
```diff
@@ -34,3 +34,22 @@ hardware_gz302_asus_lightbar_product: "18c6"
 # Flip back to 3 if a future kernel/firmware combo fixes the regression.
 # Values: 0 = use default, 1 = ignore, 2 = disable, 3 = enable.
 hardware_gz302_wifi_powersave: 2
+
+# Hibernation backing swap. BTRFS swap files require a NOCOW subvolume
+# with no compression; `btrfs filesystem mkswapfile` enforces both. Size
+# = detected RAM (memtotal_mb) rounded up to GB + 4 GB headroom for the
+# image header. memtotal_mb is the kernel-visible total, which is already
+# net of amdgpu's VRAM carveout on Strix Halo.
+hardware_gz302_swap_size: "{{ ((ansible_facts['memtotal_mb'] / 1024) | round(0, 'ceil') | int + 4) }}g"
```
palazzem: KISS: 128GB, RAM can't be expanded on that model.

roles/hardware/vars/main.yml
```diff
@@ -34,3 +34,22 @@ hardware_gz302_asus_lightbar_product: "18c6"
...
+
+# HibernateDelaySec for systemd-suspend-then-hibernate.service: time in
+# s2idle before transitioning to hibernate. 1800s = 30 min trades a small
+# battery cost (suspend draws ~1-2 %/h on Strix Halo) for fast resume on
+# short lid-close intervals.
+hardware_gz302_hibernate_delay_sec: 1800
```
palazzem: Move it to 1h

roles/hardware/vars/main.yml
```diff
@@ -34,3 +34,22 @@ hardware_gz302_asus_lightbar_product: "18c6"
...
+
+# KDE PowerDevil SleepMode enum (Plasma 6 powerdevilrc, [Profile][SuspendAndShutdown]).
+# Required because PowerDevil holds a block-mode inhibitor on
+# handle-lid-switch, making /etc/systemd/logind.conf.d/ inert under KDE.
+# Values: 1 = Suspend, 2 = Hibernate, 3 = Suspend then Hibernate.
+hardware_gz302_kde_sleep_mode: 3
```
palazzem: This is unneeded. Just hardcode in the copy block.

roles/hardware/tasks/gz302.yml
palazzem: Done in 88feedc. Collapsed both tasks into a single `swapon --all` — `--all` activates everything in /etc/fstab and silently skips devices already in use (util-linux behavior), so the precheck is no longer needed.

roles/hardware/tasks/gz302.yml
palazzem: Done in 88feedc. Replaced with a single idempotent `swapon --all` (paired with the precheck removal above).

roles/hardware/tasks/gz302.yml
palazzem: Done in 88feedc. Switched to `ansible.builtin.copy` matching the sleep/logind drop-ins, with `SleepMode=3` hardcoded under both `[AC][SuspendAndShutdown]` and `[Battery][SuspendAndShutdown]`.

One trade-off worth flagging: `copy` overwrites the entire `~/.config/powerdevilrc`, so any other PowerDevil settings KDE has written (display dim, brightness profiles, etc.) get wiped on every run. `ini_file` was surgical; `copy` makes Hanzo the sole owner of that file. Fine for a fresh provision, but worth noting since powerdevilrc is a user-managed file (unlike the systemd drop-ins which live in `.d/` directories).

roles/hardware/vars/main.yml
palazzem: Done in 88feedc. Hardcoded `hardware_gz302_swap_size: 128g` and rewrote the comment to explain the KISS rationale (soldered LPDDR5X, fixed cap covers every SKU).

roles/hardware/vars/main.yml
palazzem: Done in 88feedc. `hardware_gz302_hibernate_delay_sec: 3600`.

roles/hardware/vars/main.yml
palazzem: Done in 88feedc. Removed the `hardware_gz302_kde_sleep_mode` var; `SleepMode=3` is now inlined in the copy block.

PR #283, opened 2026-05-23
refactor(playbook): consolidate systemd detection into pre_tasks fact

### LINKED ISSUE #282: refactor(playbook): consolidate systemd detection into pre_tasks fact
## Context

`playbook.yml` computes `model_is_gz302` in `pre_tasks` using
`ansible_facts['virtualization_type']` to skip the hardware role inside
containers. The fact is unreliable: in the project's docker test container
it resolves to `'NA'` rather than `'docker'`, so the "not in [docker,
container, podman]" guard evaluates `True` and the gate fails open. Today
this is masked by the `'product_name' in ansible_facts` short-circuit (no
DMI in the container), but it's a latent footgun.

Three roles have already independently fixed this for themselves by
stat-ing `/run/systemd/system` and gating on `*_systemd_running.stat.exists`:

- `roles/hardware/tasks/gz302.yml:33-38` (`hardware_systemd_running`,
  consumed at `:48` and `:173`).
- `roles/home/tasks/main.yml:43-47` (`home_systemd_running`, consumed at
  `:62`).
- `roles/virtualization/tasks/main.yml:38-43`
  (`virtualization_systemd_running`, consumed at `:56`).

`roles/virtualization/tasks/main.yml:45-52` documents the rationale verbatim:

> Without PID-1 systemd (container builds, chroots, BuildKit RUN steps),
> libvirtd cannot start — gate on `/run/systemd/system` rather than
> `ansible_facts['virtualization_type']`, because the latter is unreliable
> under BuildKit (the docker fact is not always populated at play start,
> so the previous guard leaked the systemd_service call into container
> builds).

So: the pattern is already battle-tested locally, just duplicated three
times. This PR lifts it into `playbook.yml` `pre_tasks` as a single fact,
deduplicates the three role-local stats, and drops the broken
virtualization-type check from `model_is_gz302`. CLAUDE.md rule 6 (DRY,
extract patterns appearing 2+ times) makes the consolidation mandatory
once we touch any of these sites.

Why this PR exists *now*: a held follow-up refactor (worktree
`.worktrees/refactor-system`) moves hibernate setup into a role that runs
unconditionally and would allocate a 128 GB btrfs swapfile inside the test
container without a reliable gate. Landing detection first gives that
refactor a clean primitive to consume.

References for the `/run/systemd/system` check:
- [`sd_booted(3)` manpage](https://www.freedesktop.org/software/systemd/man/latest/sd_booted.html):
  "Internally, this function checks whether the directory
  `/run/systemd/system/` exists." This is the canonical systemd-booted
  marker.
- Background on the virtualization-type brittleness:
  [ansible/ansible#75118](https://github.com/ansible/ansible/issues/75118)
  (CentOS 7 docker reports `kvm`),
  [ansible/ansible#12573](https://github.com/ansible/ansible/issues/12573)
  (Docker-on-EC2 reports `xen`),
  [ansible/ansible#80866](https://github.com/ansible/ansible/issues/80866)
  (rootless podman not detected),
  [moby/buildkit#4503](https://github.com/moby/buildkit/issues/4503)
  (`/.dockerenv` absent during `docker build` RUN).

The exact reason `ansible_facts['virtualization_type']` returns `'NA'` in
the playbook context (but `'docker'` for an ad-hoc `setup -a
'gather_subset=virtual,!all'` against the same image) is not characterised
in this PR. It does not need to be: we're removing the fact, not fixing
it.

### New fact

Expose one boolean at the playbook level:

- `host_has_systemd` — `True` iff `/run/systemd/system` exists at play
  start.

Computed once in `pre_tasks`, tagged `[always]` so it's available under
`--tags <role>` invocations (same mechanism as the existing
`model_is_gz302` set_fact at `playbook.yml:41-46`, which already proves
the tag pattern works).

### Updated fact

Keep `model_is_gz302`. Replace its virtualization clause with
`host_has_systemd`. The product_name check stays (it's the actual hardware
discriminator; the systemd clause just ensures we're on a bootable host,
not a chroot or build container).

### Don't add `host_is_container`

No consumer needs to distinguish "container" from "non-systemd init". Every
existing site that wants to skip in containers also wants to skip on
non-systemd inits (libvirtd, asusd, keyd, user systemd units — all
no-ops without systemd PID 1). YAGNI.

### Fact organisation

CLAUDE.md rule 3 already establishes the contract: detection facts are
computed in `playbook.yml` `pre_tasks`, role tasks consume them via `when:`.
This PR preserves that contract and extends it from `model_is_gz302` to
`host_has_systemd`.

### gather_subset

Current: `['!all', 'min', 'hardware', 'virtual']`. After this PR, nothing
in the repo reads any `virtual`-subset fact. Verified with `git grep
virtualization_` across the worktree — only matches are the playbook fact
this PR removes and the virtualization-role comment block (prose, not a
read). Drop `virtual` from the list.

### 1. `playbook.yml`

Replace the gather_subset comment + `virtual` entry:

```yaml
  # Restrict fact gathering to the subsets actually consumed by the
  # roles:
  #   - min:      ansible_facts['user_id'], ansible_facts['env']
  #   - hardware: ansible_facts['product_name'] (from /sys/class/dmi/id)
  # Container/host-shape detection uses a stat on /run/systemd/system
  # in pre_tasks rather than ansible's virtualization_type fact, which
  # is unreliable under BuildKit (see commit body).
  # Skipping unused subsets (network, virtual, ohai, facter) cuts
  # startup time on every run, especially noticeable in the CI build.
  gather_subset:
    - "!all"
    - min
    - hardware
```

Replace the `Detect GZ302 hardware` set_fact with two tasks:

```yaml
    # /run/systemd/system is the canonical sd_booted(3) marker.
    # Absent during `docker build` RUN, on non-systemd inits, and in
    # chroots — exactly the contexts where the systemd-touching tasks
    # in this repo must skip.
    - name: Detect running systemd
      ansible.builtin.stat:
        path: /run/systemd/system
      register: hanzo_systemd_stat
      become: false
      tags: [always]

    - name: Compute provisioning facts
      ansible.builtin.set_fact:
        host_has_systemd: "{{ hanzo_systemd_stat.stat.exists }}"
        model_is_gz302: "{{ 'product_name' in ansible_facts
                           and 'GZ302' in ansible_facts['product_name']
                           and hanzo_systemd_stat.stat.exists }}"
      tags: [always]
```

Implementation notes:
- **No `check_mode: false`** on the stat. The three existing role-local
  stat tasks (`gz302.yml:37`, `home/tasks/main.yml:47`,
  `virtualization/tasks/main.yml:42`) cargo-cult that line, but
  `ansible.builtin.stat` is documented as fully supporting check mode
  and returns real results. Drop it here and in the three deletions
  below. **Empirical validation is mandatory**, not just documentary:
  the three stats were independently added with that line by past
  authors, so before final commit, verification step 3 below (fact-value
  inspection under `--check`) must confirm `host_has_systemd: false`
  resolves correctly inside the container *without* `check_mode: false`
  on the new stat task. If that ever returns `True` or a missing
  `.stat.exists` under `--check`, restore `check_mode: false` on the
  pre_tasks stat (and revisit whether the three role-local omissions
  are also unsafe). This is the load-bearing assumption of the
  consolidation and must not be left untested.
- **Explicit `become: false`** on the stat per CLAUDE.md rule 2 ("Every
  task declares `become:` true or false"). `set_fact` inherits the
  play-level `become: false` so no explicit declaration is needed there
  (matching the existing `Detect GZ302 hardware` task at `:41-46`).
- **`tags: [always]`** on both tasks so selective `--tags <role>` runs
  still get the facts.

### 2. `roles/hardware/tasks/gz302.yml`

Delete the local stat block (`:33-38`):

```yaml
- name: Detect running systemd
  ansible.builtin.stat:
    path: /run/systemd/system
  register: hardware_systemd_running
  check_mode: false
  become: false
```

Update the two consumers (`:48` and `:173`):

```yaml
# was:
#   - hardware_systemd_running.stat.exists
# now:
    - host_has_systemd
```

### 3. `roles/home/tasks/main.yml`

Delete the local stat block (`:43-47`):

```yaml
- name: Detect running systemd
  ansible.builtin.stat:
    path: /run/systemd/system
  register: home_systemd_running
  check_mode: false
```

Update the consumer (`:62`):

```yaml
# was:
#   - home_systemd_running.stat.exists
# now:
    - host_has_systemd
```

Leave the `XDG_RUNTIME_DIR` and `not ansible_check_mode` clauses intact —
they cover orthogonal conditions (user-session vs system-wide systemd, and
check-mode skip).

### 4. `roles/virtualization/tasks/main.yml`

Delete the local stat block (`:38-43`):

```yaml
- name: Detect running systemd
  ansible.builtin.stat:
    path: /run/systemd/system
  register: virtualization_systemd_running
  check_mode: false
  become: false
```

Rewrite the comment block at `:45-52` — the rationale it documents
(virt-type unreliable under BuildKit) is now the *primary* commit
justification, but the file-local prose should point at the shared fact:

```yaml
# --check simulates the pacman install, so virsh wouldn't exist on
# first run. Without PID-1 systemd (container builds, chroots, BuildKit
# RUN steps), libvirtd cannot start either — gate on `host_has_systemd`
# (computed in playbook.yml pre_tasks via a stat on /run/systemd/system
# — the canonical sd_booted(3) marker) rather than ansible's
# virtualization_type fact, which is unreliable under BuildKit.
```

The leading `--check` sentence is preserved from the existing block
because the `not ansible_check_mode` clause on the `when:` is gated
on virsh availability (`pacman` install is simulated under check
mode), not on systemd presence — dropping the comment would orphan
the rationale for that separate guard.

Update the consumer (`:56`):

```yaml
# was:
#   - virtualization_systemd_running.stat.exists
# now:
    - host_has_systemd
```

### 5. `CLAUDE.md`

Rule 3 currently reads:

> 3. Hardware detection belongs in `roles/hardware/tasks/main.yml`. The
> `when:` clause ANDs `ansible_facts['product_name']` against
> `ansible_facts['virtualization_type']` so the role is skipped inside
> Docker / Podman / generic containers. This replaces the previous
> `SKIP_HARDWARE_CHECK` env var — moving the check into the playbook keeps
> the policy in one place and prevents host-DMI bleed-through in container
> tests.

Replace with (tightened per "minimal docs, no decoration" rule —
the `sd_booted(3)` rationale and BuildKit context belong on the
pre_tasks stat block in `playbook.yml`, where someone editing the
detection logic will read it, not in the CLAUDE.md rule itself):

> 3. Hardware/host detection is computed in `playbook.yml` `pre_tasks`
> (`host_has_systemd` from a stat on `/run/systemd/system`;
> `model_is_gz302` from `ansible_facts['product_name']` AND
> `host_has_systemd`) and consumed via `when:` on role tasks (e.g.
> `roles/hardware/tasks/main.yml:4` gates on `model_is_gz302`). Roles
> consume these facts; they never redefine them.

### 6. `README.md`

Line 70 currently reads:

> The `hardware` role is dispatched by `ansible_product_name` and skipped
> automatically inside containers (via `ansible_virtualization_type`).

Replace with:

> The `hardware` role is dispatched by `ansible_facts['product_name']` and
> skipped automatically inside containers and other non-systemd contexts
> (via a stat of `/run/systemd/system` — see `CLAUDE.md` rule 3).

## Verification

1. **Lint:** `pre-commit run --all-files` clean. `hanzo_systemd_stat`
   uses the project's playbook-level `hanzo_*` naming convention (matches
   `hanzo_aur_sudoers`, `hanzo_config_stat`); ansible-lint's
   `var-naming[no-role-prefix]` rule only fires on role-level vars and
   doesn't apply here.

2. **Container test (default, --check --diff):**
   ```
   ```
   `failed=0`. Tasks `Enable and start ASUS services` (gz302.yml),
   `Enable and start keyd` (gz302.yml), `Enable and start user services`
   (home), and `Configure libvirt default network autostart`
   (virtualization) all skip because `host_has_systemd` is `False`.

3. **Fact-value inspection (verbose, fast iteration) — proves the
   `check_mode: false` drop is safe:**
   ```
   docker build --no-cache --build-arg ANSIBLE_ARGS="--tags hardware --check --diff -v" -f tests/Containerfile -t hanzo:test .
   ```
   The `Compute provisioning facts` task must print:
   `{"ansible_facts": {"host_has_systemd": false, "model_is_gz302": false}, "changed": false}`.
   The stat task above it must show real `stat.exists: false`, not a
   skipped/check-mode placeholder. This is the empirical check that the
   "ansible.builtin.stat runs fully in check mode" docs claim holds in
   our environment despite three past authors having added
   `check_mode: false`. If the value comes back `True`, missing, or
   marked as a check-mode no-op, the consolidation is unsafe as
   specified and `check_mode: false` must be restored on the new
   pre_tasks stat. The existing CI only runs in check mode, so no
   separate `ANSIBLE_ARGS=""` run is required — and isn't realistic
   anyway (it would hit unrelated pacman / btrfs / limine paths that
   have nothing to do with detection). The 128 GB swapfile concern is a
   property of the *held refactor* PR's tests, not this one.

4. **Grep audits** (must all return empty):
   ```
   git grep ansible_virtualization_type
   git grep ansible_facts\\[\\'virtualization_type\\'\\]
   git grep hardware_systemd_running
   git grep home_systemd_running
   git grep virtualization_systemd_running
   ```

5. **Real-host verification is intentionally out of scope.** CLAUDE.md
   prohibits running `ansible-playbook` on the host, and no CI runs on
   real hardware. The change is safe on the real GZ302 by construction:
   `/run/systemd/system` exists on every booted systemd host. If the
   reviewer wants a stronger check, the only path is manual: clone the
   branch onto a GZ302, run `stat /run/systemd/system` once (a
   non-mutating read), and confirm `.stat.exists` is `True`.

## Out of scope

- **Don't move hibernate.** That's the held refactor PR
  (`.worktrees/refactor-system`). Hibernate stays in
  `roles/hardware/tasks/gz302.yml` for this PR.
- **Don't rename `foundation` → `system`.** Held refactor.
- **Don't add `host_is_container`.** No consumer; YAGNI.
- **Don't add `community.docker.current_container_facts`.** Another
  collection install for a problem a 2-line stat already solves.
- **Don't touch `community.general` / `kewlfft.aur` versions** in
  `requirements.yml`.
- **Don't lift any *other* repeated detection pattern** beyond the three
  systemd stats explicitly listed. If `git grep` surfaces another
  candidate, note it for a follow-up — don't expand this PR's diff.

## Held-PR coupling

The held refactor branch (`.worktrees/refactor-system`,
`refactor/system-role-and-hibernate-move`) does the following to the
files this PR touches:

- **`roles/hardware/tasks/gz302.yml`**: removes the hibernate block
  (lines 74-151 of current main) but **keeps** the
  `hardware_systemd_running` stat and its two consumers
  (`Enable and start ASUS services`, `Enable and start keyd`).
- **`playbook.yml`**: renames the foundation role import to `system`;
  leaves the `Detect GZ302 hardware` set_fact unchanged.
- **`README.md`** and **`CLAUDE.md`**: untouched.

When the refactor rebases on top of this detection PR:
- The two `hardware_systemd_running.stat.exists` lines in gz302.yml will
  no longer exist (this PR removed them). The refactor's gz302.yml diff
  needs to *not* re-introduce them — `git rebase` will surface this as
  a conflict on those `when:` clauses; resolution is to take the
  detection PR's `host_has_systemd` form.
- The playbook rename is independent (just `name: foundation` →
  `name: system`) and won't conflict with the pre_tasks edits here.

This is a normal rebase concern, not a blocker for either PR. Land
detection first, then resume the refactor.

## Definition of done

- [ ] `playbook.yml`: `host_has_systemd` and `model_is_gz302` set via
      pre_tasks, computed from a single stat of `/run/systemd/system`.
- [ ] `playbook.yml`: `gather_subset` no longer includes `virtual`; the
      comment block above it is updated.
- [ ] `roles/hardware/tasks/gz302.yml`: local stat block removed; two
      `when:` clauses use `host_has_systemd`.
- [ ] `roles/home/tasks/main.yml`: local stat block removed; one
      `when:` clause uses `host_has_systemd`.
- [ ] `roles/virtualization/tasks/main.yml`: local stat block removed;
      one `when:` clause uses `host_has_systemd`; comment block at the
      former `:45-52` rewritten to point at the playbook-level fact.
- [ ] `CLAUDE.md` rule 3 reflects the new detection mechanism while
      preserving the "gate lives on the role include, fact lives in
      pre_tasks" split.
- [ ] `README.md` line 70 updated.
- [ ] `pre-commit run --all-files` clean.
- [ ] `docker build -f tests/Containerfile -t hanzo:test .` passes
      (`failed=0`) under both `--check` (default) and `ANSIBLE_ARGS=""`.
- [ ] All five `git grep` audits return empty.
- [ ] Conventional Commits message, e.g.
      `refactor(playbook): consolidate systemd detection into pre_tasks fact`.

palazzem:
### Related Issues

- Closes #282

### Proposed Changes:

Consolidate three role-local `stat /run/systemd/system` blocks (`hardware/gz302`, `home`, `virtualization`) into a single playbook-level `host_has_systemd` fact computed once in `playbook.yml` `pre_tasks`. Roles now consume the shared fact via `when:` rather than each re-running its own stat.

Also drops the `ansible_facts['virtualization_type']` clause from `model_is_gz302`. That fact returns `'NA'` under BuildKit despite running inside a docker container, so the previous container guard fell open — only the missing DMI on container hosts saved it from misfiring. `/run/systemd/system` is the canonical `sd_booted(3)` marker and is absent in exactly the contexts where the systemd-touching tasks must skip (`docker build` RUN, non-systemd inits, chroots), so it's the correct primitive.

`gather_subset` drops `virtual` since no consumer reads any virt-subset fact afterward.

Net -15 lines across six files (`playbook.yml`, `roles/hardware/tasks/gz302.yml`, `roles/home/tasks/main.yml`, `roles/virtualization/tasks/main.yml`, `CLAUDE.md`, `README.md`).

### Testing:

- `pre-commit run --all-files` passes
- `docker build -f tests/Containerfile -t hanzo:test .` returns `failed=0` (PLAY RECAP: `ok=57 changed=30 unreachable=0 failed=0 skipped=26`)
- Empirically validated under `--tags hardware --check -v`: `Detect running systemd` returns real `stat.exists: false` and `Compute provisioning facts` returns `{"host_has_systemd": false, "model_is_gz302": false}` — proves dropping `check_mode: false` from the stat is safe
- Audit greps for all five removed identifiers (`ansible_virtualization_type`, `ansible_facts['virtualization_type']`, `hardware_systemd_running`, `home_systemd_running`, `virtualization_systemd_running`) return empty

### Extra Notes (optional):

The `check_mode: false` on the prior role-local stats was cargo-culted from when the stat needed to bypass `--check`; the consolidated stat doesn't need it (verified above). Detection rationale now lives in a single commented block in `playbook.yml` pre_tasks; `CLAUDE.md` rule 3 and the README hardware paragraph were updated to reflect the new structure.

### Checklist


roles/home/tasks/main.yml
```diff
@@ -40,16 +40,10 @@
     git_repo: "{{ home_claude_config_repo }}"
     git_dest: "{{ home_claude_config_dir | expanduser }}"
 
-- name: Detect running systemd
-  ansible.builtin.stat:
-    path: /run/systemd/system
-  register: home_systemd_running
-  check_mode: false
-
 # User-scope systemd also needs an active session bus (XDG_RUNTIME_DIR
```
palazzem: Remove the entire comment as it is unneeded.

roles/virtualization/tasks/main.yml
```diff
@@ -35,25 +35,14 @@
...
-
-# --check simulates the pacman install, so virsh wouldn't exist on
-# first run. Without PID-1 systemd (container builds, chroots, BuildKit
-# RUN steps), libvirtd cannot start — gate on /run/systemd/system rather
-# than ansible_facts['virtualization_type'], because the latter is
-# unreliable under BuildKit (the docker fact is not always populated at
-# play start, so the previous guard leaked the systemd_service call
-# into container builds). Same gating primitive as roles/home and
-# roles/hardware/gz302.yml.
+# Two conditions skip this block: --check simulates the pacman install,
```
palazzem: Remove, comment unneeded.

CLAUDE.md
```diff
@@ -6,7 +6,7 @@ CachyOS system provisioner powered by Ansible.
...
 2. Explicit `become:` on every task. The playbook default is `become: false`. Every task declares `become: true` (system operations: package installs, service management, config writes) or `become: false` (user-space: paru, dotfiles, language version managers).
-3. Hardware detection belongs in `roles/hardware/tasks/main.yml`. The `when:` clause ANDs `ansible_facts['product_name']` against `ansible_facts['virtualization_type']` so the role is skipped inside Docker / Podman / generic containers. This replaces the previous `SKIP_HARDWARE_CHECK` env var — moving the check into the playbook keeps the policy in one place and prevents host-DMI bleed-through in container tests.
+3. Host-shape facts (`host_has_systemd`, `model_is_gz302`) are computed once in `playbook.yml` `pre_tasks` and consumed via `when:` on role tasks. Roles never redefine them. See the pre_tasks block for the detection rationale.
```
palazzem: This will get obsolete fast. The rule should be about having system checks to skip tasks in the playbook.yml, not what they are, how to use it etc... coding style is defined by real code, here you must put the rule. Rewrite this to be a generic rule without examples of what has been already implemented.

playbook.yml
palazzem: Remove entirely the comment as it is unneeded.

playbook.yml
```diff
@@ -27,22 +25,29 @@
...
-    # so a pure DMI check would mis-trigger inside the container. We
-    # AND in ansible_facts['virtualization_type'] to ensure we never
-    # write to host paths from inside docker/podman.
-    #
-    # `'product_name' in ansible_facts` short-circuits the second
-    # condition so the `'GZ302' in ...` check is safe even when the
-    # DMI fact is missing (e.g., when `gather_subset` doesn't include
-    # `hardware`).
-    - name: Detect GZ302 hardware
+    # /run/systemd/system is the canonical sd_booted(3) marker. Absent
```
palazzem: Bad comment. Rewrite only this part:
```
# /run/systemd/system is the canonical sd_booted(3) marker. Absent
    # during `docker build` RUN, on non-systemd inits, and in chroots —
    # exactly the contexts where the systemd-touching tasks in this
    # repo must skip.
```

It must state Why we use systemd check, you are adding an informal touch like you are talking to somebody. This is an engineering comment, keep it short and clear.

roles/home/tasks/main.yml
palazzem: Removed the entire XDG comment block. The `scope: user` line on the task and the `XDG_RUNTIME_DIR` clause in the `when:` carry the intent without prose.

roles/virtualization/tasks/main.yml
palazzem: Removed.

CLAUDE.md
palazzem: Rewrote rule 3 as a generic rule: `Detection facts that gate task execution are defined in `playbook.yml` pre_tasks; roles consume them via `when:` and never redefine them.` No fact names, no examples, no file:line references — the rule lives on its own and the actual facts speak for themselves in code.

playbook.yml
palazzem: Removed.

playbook.yml
palazzem: Rewrote that part to: `Stat-gate systemd-touching tasks on the sd_booted(3) marker /run/systemd/system — absent under \`docker build\` RUN, chroots, and non-systemd inits.` Leads with the action, states why we gate (sd_booted marker semantics), drops the informal connector. The virtualization_type sentence stays untouched per your callout.

PR #284, opened 2026-05-23
refactor(system): `foundation` to `system` role; relocate hibernate and devtools

palazzem:
### Related Issues

Follow-up to #283 (`host_has_systemd` primitive). No tracking issue.

### Proposed Changes:

- **`roles/foundation/` → `roles/system/`.** Rename the role directory, rename `foundation_*` vars to `system_*`, and update the playbook role import and tag accordingly.
- **Move hibernate setup out of `roles/hardware/tasks/gz302.yml` into the new `roles/system/tasks/hibernate.yml`.** Hibernate is not a GZ302-specific concern — the swap subvolume, `mkswapfile`, fstab entry, `swapon`, `HibernateDelaySec` drop-in, logind lid handler, and KDE PowerDevil `SleepMode=3` all apply to any bootable systemd target. The include is gated `when: host_has_systemd` so `docker build` RUN contexts (no systemd PID 1) skip the entire block — without this gate the 128 GB btrfs swapfile would be allocated inside the test container. The `Reload logind` handler moves with its sole notifier from `roles/hardware/handlers/main.yml` to a new `roles/system/handlers/main.yml`. `system_swap_size` and `system_hibernate_delay_sec` move from hardware vars to system vars (the GZ302 max-RAM sizing rationale stays as inline comment documentation).
- **Split `github-cli` out of `system_packages` baseline.** Move it into a new `roles/tools/tasks/devtools.yml` backed by `tools_dev_packages: [github-cli]`, included from `roles/tools/tasks/main.yml`. Matches the existing per-domain pattern in `tools/` (`python_tools.yml`, `node_tools.yml`, `iac.yml`). Keeps `system_packages` to true baseline CLI utilities and gives `tools_dev_packages` room to grow as additional development tools land.

### Testing:

- `pre-commit run --all-files` passes
- `docker build -f tests/Containerfile -t hanzo:test .` returns `failed=0` (`PLAY RECAP: ok=59 changed=31 unreachable=0 failed=0 skipped=27`)
- Hibernate guard empirically verified: `TASK [system : Configure hibernation]` shows `skipping: [localhost]` in the container build, confirming no nested btrfs / swapfile / HibernateDelaySec tasks fire in the no-systemd context

### Extra Notes (optional):

- KDE PowerDevil `SleepMode=3` is intentionally co-located with the rest of the hibernate setup (rather than split into `desktop`/`home`) because it only matters as part of the suspend-then-hibernate policy this file establishes. Co-location keeps the feature reviewable as a unit.
- `tools_dev_packages` ships with a single entry today; the list shape is the extensibility seam for future developer-CLI additions.
- Net +14 lines (131 insertions, 117 deletions across 12 files).

### Checklist


roles/system/vars/main.yml
```diff
@@ -0,0 +1,25 @@
...
+# Hibernation backing swap. BTRFS swap files require a NOCOW subvolume
+# with no compression; `btrfs filesystem mkswapfile` enforces both. Sized
+# at the GZ302's max RAM SKU (128 GB) — RAM is soldered LPDDR5X, so a
+# fixed cap covers every variant without per-host math.
+system_swap_size: 128g
+
+# HibernateDelaySec for systemd-suspend-then-hibernate.service: time in
+# s2idle before transitioning to hibernate. 1h gives a long fast-resume
+# window before paying the disk-write cost.
+system_hibernate_delay_sec: 3600
```
palazzem: Move this to 3h

roles/system/vars/main.yml
```diff
@@ -0,0 +1,25 @@
...
+  - wget
+  - jq
+  - yq
+  - httpie
+  - whois
+  - tmux
+
+system_locale: en_US.UTF-8
+
+# Hibernation backing swap. BTRFS swap files require a NOCOW subvolume
```
palazzem: This comment is unneeded. You are explaining details unrelated to the swap size. Just mention Hibernation for a 128GB system.

roles/system/vars/main.yml
palazzem: Bumped `system_hibernate_delay_sec` from 3600 → 10800 (3h). Updated the inline comment from "1h" → "3h" so it stays accurate.

roles/system/vars/main.yml
palazzem: Trimmed to a one-liner: `# Hibernation swap for a 128 GB system.` Dropped the BTRFS/NOCOW/LPDDR5X details — they're implementation/hardware context, not about the size.

Issue #290, opened 2026-05-30
Hibernation doesn't work


palazzem:
We should include these steps in the hibernation role: https://wiki.cachyos.org/configuration/general_system_tweaks/#swap-configuration

palazzem:
**Claude Harness**

Run intake for this issue (type: **bug**, fast lane). Completeness details confirmed with the user:

**Expected behavior**
After provisioning, hibernation works: the machine powers off on hibernate and restores the saved session on next boot.

**Actual behavior**
Hibernate fails: the machine does not power back on cleanly — after invoking hibernate it doesn't turn on, and after some reboots it starts fresh with the session lost.

**Reproduction path**
Provision with the current `system` role (hibernate tasks), then trigger hibernation (e.g. `systemctl hibernate` or suspend-then-hibernate via lid close) and power the machine back on.

**Scope ruling (user-confirmed)**
Adopt the **full** "Switching from ZRam to Zswap" procedure from https://wiki.cachyos.org/configuration/general_system_tweaks/#swap-configuration — not just a hibernation-minimum subset. The current `roles/system/tasks/hibernate.yml` covers only the Btrfs swap subvolume/swapfile/fstab/swapon steps. Missing wiki steps to add:

1. Disable ZRam via kernel parameter `systemd.zram=0`.
2. Enable Zswap via kernel parameters `zswap.enabled=1 zswap.shrinker_enabled=1 zswap.compressor=lz4 zswap.max_pool_percent=30`.
3. Create an empty override `/etc/udev/rules.d/30-zram.rules` to neutralize CachyOS's udev rule that disables Zswap.
4. Add `lz4` to the `MODULES` array in `/etc/mkinitcpio.conf`.
5. Rebuild the initramfs.

**Environment facts**
- Filesystem: Btrfs (the role already uses `btrfs filesystem mkswapfile`).
- Bootloader: Limine — the repo already manages kernel params via `/etc/default/limine` (`roles/hardware/tasks/gz302.yml`) and has a `limine-mkinitcpio` handler; the initramfs rebuild step is `limine-mkinitcpio`.

palazzem:
Fixed in #291

Issue #243, opened 2026-05-14
Improve container test coverage beyond --check structural validation


palazzem:
## Context

The Ansible migration in #241 / #242 ships with a `tests/Containerfile` that runs a two-stage `ansible-playbook --check --diff` (with and without user config). This mirrors the previous `pyinfra --dry` behaviour — coverage is at structural-validation parity with what we had before, not a regression introduced by the migration.

A self-review pass during #242 surfaced that `--check` mode has real limits for this codebase:

- **`git_adopt.yml` branches are gated by `not ansible_check_mode`.** In `--check` mode only the upstream `stat` probes run; the actual pull/clone/adopt git operations are skipped. So Stage 1 vs Stage 2 differ only in which `when:` branches evaluate true, not in which logic runs end-to-end.
- **The "adopt" branch (which does `git reset --hard origin/main`) is never exercised.** Stage 2 of the container test creates `.dotfiles/.git` and `.claude/.git` as empty directories, so the adopt path can never trigger.
- **No idempotency check.** The canonical Ansible test pattern — "run twice; second run reports `changed=0`" — is absent. `--check` mode makes it impossible to assert (everything reports "would change").
- **Pass/fail signal is only the playbook exit code.** No assertion that expected tasks ran or were skipped for the *right* reason. A future regression that weakens the hardware-role container guard, or that reorders task gating, would not be caught.
- **`roles/devtools`, `roles/languages`, `roles/infra` are effectively unexercised in check mode** because every task with side effects is gated on `not ansible_check_mode`.

These limits are a consequence of the security rule in `CLAUDE.md` ("NEVER run `ansible-playbook` on the host machine"), which forces all real validation into the container.

## Proposed improvements (not prioritised — pick what makes sense)

1. **Add `ansible.builtin.assert` tasks at strategic points** (or in a CI-only role appended in the Containerfile) that verify invariants under check mode. Examples: `dotfiles_git_dir.stat.exists == True` in Stage 2 / `False` in Stage 1; `hanzo_fullname is defined` in Stage 2 / `is not defined` in Stage 1; `ansible_virtualization_type in ['docker','container','podman']` so the hardware-role guard is positively asserted.

2. **Seed real local git repos inside the container** (`git init --bare /tmp/fake-dotfiles.git`, populated with at least one commit on `main`) and override `dotfiles_repo` via `--extra-vars`. Then drop the `not ansible_check_mode` gates on the git_adopt branches so all three paths (pull, adopt, clone) actually run.

[...]

PR #291, opened 2026-08-17
fix(system): switch ZRam to Zswap so hibernation resumes

palazzem:
**Claude Harness**

Hibernation never resumed because provisioning left CachyOS's ZRam setup in place: `zram-generator` declares a RAM-sized `zram0` swap at priority 100 and `cachyos-settings` ships a udev rule that force-disables Zswap on every `zram0` change event, so the hibernation image had no persistent backing store. The hibernation tasks previously implemented only the Btrfs swapfile half of the CachyOS wiki "Switching from ZRam to Zswap" procedure; this PR adds the missing steps — `systemd.zram=0` and the `zswap.*` kernel params in `/etc/default/limine`, a rule-free `/etc/udev/rules.d/30-zram.rules` override, `lz4` appended to the initramfs `MODULES` via an `/etc/mkinitcpio.conf.d` drop-in, and a `limine-mkinitcpio` rebuild handler.

## Run record

- **Issue:** #290
- **Type:** bug — confirmed at `/deliver`
- **Session:** `7f4e4da2-18b0-4571-aa59-582bebf8fe64` — resume with `claude --resume 7f4e4da2-18b0-4571-aa59-582bebf8fe64`

### DIFF
--- a/playbook.yml
+++ b/playbook.yml
@@ -138,17 +138,8 @@
           become: true
           tags: [always]
 
-  # Cross-role handlers live in the play. Handler scope is play-wide
-  # regardless of where a handler is defined, and ansible dedupes handler
-  # runs by name — so two roles notifying one play-level handler produce a
-  # single run, while two identically-bodied role-local handlers under
-  # different names would run twice. limine-mkinitcpio regenerates the
-  # initramfs for every installed kernel and rewrites the Limine boot
-  # entries from /etc/default/limine; both the hardware role (GZ302 GPU
-  # modprobe config, OLED kernel param) and the system role (Zswap kernel
-  # params, lz4 initramfs module) change inputs to that same command.
-  # Notified handlers run regardless of --tags filtering, so no tag is
-  # needed here for selective runs (`--tags hardware`, `--tags system`).
+  # Regenerates the initramfs for every installed kernel and rewrites the
+  # Limine boot entries from /etc/default/limine.
   handlers:
     - name: Regenerate initramfs
       ansible.builtin.command: limine-mkinitcpio
--- a/playbook.yml
+++ b/playbook.yml
@@ -137,3 +137,20 @@
             state: absent
           become: true
           tags: [always]
+
+  # Cross-role handlers live in the play. Handler scope is play-wide
+  # regardless of where a handler is defined, and ansible dedupes handler
+  # runs by name — so two roles notifying one play-level handler produce a
+  # single run, while two identically-bodied role-local handlers under
+  # different names would run twice. limine-mkinitcpio regenerates the
+  # initramfs for every installed kernel and rewrites the Limine boot
+  # entries from /etc/default/limine; both the hardware role (GZ302 GPU
+  # modprobe config, OLED kernel param) and the system role (Zswap kernel
+  # params, lz4 initramfs module) change inputs to that same command.
+  # Notified handlers run regardless of --tags filtering, so no tag is
+  # needed here for selective runs (`--tags hardware`, `--tags system`).
+  handlers:
+    - name: Regenerate initramfs
+      ansible.builtin.command: limine-mkinitcpio
+      changed_when: true
+      become: true
--- a/roles/hardware/handlers/main.yml
+++ b/roles/hardware/handlers/main.yml
@@ -9,10 +9,9 @@
   changed_when: true
   become: true
 
-- name: Regenerate initramfs
-  ansible.builtin.command: limine-mkinitcpio
-  changed_when: true
-  become: true
+# `Regenerate initramfs`, notified by the GZ302 tasks in this role, is
+# defined at play level in playbook.yml: the system role notifies it too,
+# and one play-level definition keeps ansible's per-name dedupe intact.
 
 - name: Restart keyd
   ansible.builtin.systemd_service:
--- a/roles/hardware/handlers/main.yml
+++ b/roles/hardware/handlers/main.yml
@@ -9,10 +9,6 @@
   changed_when: true
   become: true
 
-# `Regenerate initramfs`, notified by the GZ302 tasks in this role, is
-# defined at play level in playbook.yml: the system role notifies it too,
-# and one play-level definition keeps ansible's per-name dedupe intact.
-
 - name: Restart keyd
   ansible.builtin.systemd_service:
     name: keyd
--- a/roles/system/handlers/main.yml
+++ b/roles/system/handlers/main.yml
@@ -1,6 +1,16 @@
 ---
 # logind reads its config on SIGHUP; a full restart would terminate
 # every active login session. Supported on systemd v243+.
+# Deliberately role-local rather than reusing the hardware role's
+# equivalent handler: handler names resolve globally, and sharing one
+# would tie the system role's hibernation tasks to a role that only runs
+# on matching hardware. limine-mkinitcpio regenerates the initramfs and
+# rewrites the Limine boot entries from /etc/default/limine.
+- name: Rebuild initramfs
+  ansible.builtin.command: limine-mkinitcpio
+  changed_when: true
+  become: true
+
 - name: Reload logind
   ansible.builtin.command: systemctl kill --signal=HUP systemd-logind
   changed_when: true
--- a/roles/system/handlers/main.yml
+++ b/roles/system/handlers/main.yml
@@ -1,8 +1,4 @@
 ---
-# `Regenerate initramfs`, notified by the hibernation tasks in this role,
-# is defined at play level in playbook.yml: the hardware role notifies it
-# too, and one play-level definition keeps ansible's per-name dedupe intact.
-
 # logind reads its config on SIGHUP; a full restart would terminate
 # every active login session. Supported on systemd v243+.
 - name: Reload logind
--- a/roles/system/handlers/main.yml
+++ b/roles/system/handlers/main.yml
@@ -1,16 +1,10 @@
 ---
+# `Regenerate initramfs`, notified by the hibernation tasks in this role,
+# is defined at play level in playbook.yml: the hardware role notifies it
+# too, and one play-level definition keeps ansible's per-name dedupe intact.
+
 # logind reads its config on SIGHUP; a full restart would terminate
 # every active login session. Supported on systemd v243+.
-# Deliberately role-local rather than reusing the hardware role's
-# equivalent handler: handler names resolve globally, and sharing one
-# would tie the system role's hibernation tasks to a role that only runs
-# on matching hardware. limine-mkinitcpio regenerates the initramfs and
-# rewrites the Limine boot entries from /etc/default/limine.
-- name: Rebuild initramfs
-  ansible.builtin.command: limine-mkinitcpio
-  changed_when: true
-  become: true
-
 - name: Reload logind
   ansible.builtin.command: systemctl kill --signal=HUP systemd-logind
   changed_when: true
--- a/roles/system/tasks/hibernate.yml
+++ b/roles/system/tasks/hibernate.yml
@@ -76,14 +76,9 @@
 # mounted, so lz4 must ship inside the initramfs. mkinitcpio (>= 38)
 # concatenates /etc/mkinitcpio.conf.d/*.conf after the main config and
 # sources the result, so `MODULES+=(lz4)` appends to the MODULES array
-# without rewriting /etc/mkinitcpio.conf in place.
-- name: Ensure mkinitcpio drop-in directory exists
-  ansible.builtin.file:
-    path: /etc/mkinitcpio.conf.d
-    state: directory
-    mode: "0755"
-  become: true
-
+# without rewriting /etc/mkinitcpio.conf in place. The mkinitcpio package
+# owns /etc/mkinitcpio.conf.d/, so — unlike the systemd drop-in
+# directories below — it needs no pre-create task.
 - name: Add lz4 to initramfs MODULES
   ansible.builtin.copy:
     dest: /etc/mkinitcpio.conf.d/10-zswap.conf
--- a/roles/system/tasks/hibernate.yml
+++ b/roles/system/tasks/hibernate.yml
@@ -1,4 +1,46 @@
 ---
+# CachyOS defaults to ZRam: zram-generator declares a RAM-sized zram0 swap
+# at priority 100 and cachyos-settings ships a udev rule that force-disables
+# Zswap on every zram0 change event. RAM-backed swap cannot hold a
+# hibernation image, so the machine never resumes. The tasks below follow
+# the CachyOS wiki "Switching from ZRam to Zswap" procedure:
+# https://wiki.cachyos.org/configuration/general_system_tweaks/#swap-configuration
+
+# Kernel params live in /etc/default/limine; limine-mkinitcpio regenerates
+# the boot entries and the initramfs from it, so both param tasks and the
+# mkinitcpio drop-in below share the same handler.
+- name: Ensure ZRam is disabled in Limine cmdline
+  ansible.builtin.lineinfile:
+    path: /etc/default/limine
+    regexp: '^KERNEL_CMDLINE\[default\]\+=" systemd\.zram='
+    line: 'KERNEL_CMDLINE[default]+=" {{ system_zram_kernel_param }}"'
+  become: true
+  notify: Rebuild initramfs
+
+- name: Ensure Zswap kernel params in Limine cmdline
+  ansible.builtin.lineinfile:
+    path: /etc/default/limine
+    regexp: '^KERNEL_CMDLINE\[default\]\+=" zswap\.'
+    line: 'KERNEL_CMDLINE[default]+=" {{ system_zswap_kernel_params }}"'
+  become: true
+  notify: Rebuild initramfs
+
+# /usr/lib/udev/rules.d/30-zram.rules (cachyos-settings) writes `N` to
+# /sys/module/zswap/parameters/enabled whenever zram0 changes state. A
+# rule-free file of the same name in /etc/udev/rules.d shadows it — udev
+# resolves rule files by name, /etc winning over /usr/lib.
+- name: Override CachyOS udev rule that disables Zswap
+  ansible.builtin.copy:
+    dest: /etc/udev/rules.d/30-zram.rules
+    content: |
+      # Managed by Hanzo. Do not edit manually.
+      # Intentionally rule-free: shadows /usr/lib/udev/rules.d/30-zram.rules,
+      # which disables Zswap and raises vm.swappiness for ZRam.
+    owner: root
+    group: root
+    mode: "0644"
+  become: true
+
 # Hibernation backing store. Single contiguous swap file on a dedicated
 # NOCOW subvolume (CachyOS recommended layout). The kernel reads the
 # image back via a direct device offset systemd records in the
@@ -29,6 +71,30 @@
   ansible.builtin.command: swapon --all
   changed_when: false
   become: true
+
+# zswap.compressor=lz4 is applied at boot, before the root filesystem is
+# mounted, so lz4 must ship inside the initramfs. mkinitcpio (>= 38)
+# concatenates /etc/mkinitcpio.conf.d/*.conf after the main config and
+# sources the result, so `MODULES+=(lz4)` appends to the MODULES array
+# without rewriting /etc/mkinitcpio.conf in place.
+- name: Ensure mkinitcpio drop-in directory exists
+  ansible.builtin.file:
+    path: /etc/mkinitcpio.conf.d
+    state: directory
+    mode: "0755"
+  become: true
+
+- name: Add lz4 to initramfs MODULES
+  ansible.builtin.copy:
+    dest: /etc/mkinitcpio.conf.d/10-zswap.conf
+    content: |
+      # Managed by Hanzo. Do not edit manually.
+      MODULES+=(lz4)
+    owner: root
+    group: root
+    mode: "0644"
+  become: true
+  notify: Rebuild initramfs
 
 # systemd ships sleep.conf and logind.conf, but not their .conf.d/
 # drop-in directories — ansible.builtin.copy does not create parent
--- a/roles/system/tasks/hibernate.yml
+++ b/roles/system/tasks/hibernate.yml
@@ -15,7 +15,7 @@
     regexp: '^KERNEL_CMDLINE\[default\]\+=" systemd\.zram='
     line: 'KERNEL_CMDLINE[default]+=" {{ system_zram_kernel_param }}"'
   become: true
-  notify: Rebuild initramfs
+  notify: Regenerate initramfs
 
 - name: Ensure Zswap kernel params in Limine cmdline
   ansible.builtin.lineinfile:
@@ -23,7 +23,7 @@
     regexp: '^KERNEL_CMDLINE\[default\]\+=" zswap\.'
     line: 'KERNEL_CMDLINE[default]+=" {{ system_zswap_kernel_params }}"'
   become: true
-  notify: Rebuild initramfs
+  notify: Regenerate initramfs
 
 # /usr/lib/udev/rules.d/30-zram.rules (cachyos-settings) writes `N` to
 # /sys/module/zswap/parameters/enabled whenever zram0 changes state. A
@@ -89,7 +89,7 @@
     group: root
     mode: "0644"
   become: true
-  notify: Rebuild initramfs
+  notify: Regenerate initramfs
 
 # systemd ships sleep.conf and logind.conf, but not their .conf.d/
 # drop-in directories — ansible.builtin.copy does not create parent
--- a/roles/system/vars/main.yml
+++ b/roles/system/vars/main.yml
@@ -16,6 +16,15 @@ system_locale: en_US.UTF-8
 # Hibernation swap for a 128 GB system.
 system_swap_size: 128g
 
+# Kernel params for the CachyOS "Switching from ZRam to Zswap" procedure.
+# ZRam is RAM-backed and cannot hold a hibernation image, so it is turned
+# off and Zswap takes over as the compressed cache in front of the swap
+# file. Pool percentage and compressor are the values the wiki recommends.
+system_zram_kernel_param: systemd.zram=0
+system_zswap_kernel_params: >-
+  zswap.enabled=1 zswap.shrinker_enabled=1
+  zswap.compressor=lz4 zswap.max_pool_percent=30
+
 # HibernateDelaySec for systemd-suspend-then-hibernate.service: time in
 # s2idle before transitioning to hibernate. 3h gives a long fast-resume
 # window before paying the disk-write cost.


palazzem:
**Claude Harness**

## Reproduction evidence (before the fix)

This repo has no unit-test layer for role logic, and the container test skips `roles/system/tasks/hibernate.yml` entirely (it is gated on `host_has_systemd`, false under `docker build`). Standing up a test harness that could assert provisioned hibernation state would be new test infrastructure — outside this fast-lane bug's bounds — so the reproduction is documented as commands + output against the same base image the test container uses.

### 1. CachyOS default state: ZRam active, Zswap force-disabled

```
$ docker run --rm --platform linux/amd64 cachyos/cachyos:latest sh -lc '
    pacman -Sy --noconfirm >/dev/null 2>&1
    pacman -S --noconfirm cachyos-settings mkinitcpio >/dev/null 2>&1
    cat /usr/lib/systemd/zram-generator.conf
    grep -v "^#" /usr/lib/udev/rules.d/30-zram.rules | grep .
    ls /etc/udev/rules.d/30-zram.rules
    grep "^MODULES" /etc/mkinitcpio.conf'

[zram0]
compression-algorithm = zstd
zram-size = ram
swap-priority = 100
fs-type = swap

ACTION=="change", KERNEL=="zram0", ATTR{initstate}=="1", SYSCTL{vm.swappiness}="150", \
    RUN+="/bin/sh -c 'echo N > /sys/module/zswap/parameters/enabled'"

ls: cannot access '/etc/udev/rules.d/30-zram.rules': No such file or directory
MODULES=()
```

A RAM-sized `zram0` swap sits at priority 100 (the provisioned `/swap/swapfile` gets the default `-2`), the shipped udev rule writes `N` to `/sys/module/zswap/parameters/enabled` on every `zram0` change event, there is no `/etc` override shadowing it, and `MODULES` carries no `lz4`.

### 2. The provisioner never touches any of it

```
$ git grep -niE 'zswap|systemd\.zram|30-zram|mkinitcpio\.conf|lz4' -- roles playbook.yml
$ echo "exit=$?"
exit=1
```

Zero matches at `98aa1ea`: of the wiki's "Switching from ZRam to Zswap" procedure, `roles/system/tasks/hibernate.yml` implemented only the Btrfs subvolume/swapfile/fstab/`swapon` steps. Steps 1 (`systemd.zram=0`), 2 (`zswap.*` params), 3 (`/etc/udev/rules.d/30-zram.rules` override), 5 (`lz4` in `MODULES`) and 6 (initramfs rebuild) were absent — hibernation was left with ZRam as the highest-priority swap and Zswap actively disabled, which is the failure reported in #290.

## Verification after the fix

`pre-commit run --all-files` — all hooks pass. `docker build -f tests/Containerfile -t hanzo:test .` — passes (`ok=58 changed=30 failed=0`), though it does not exercise the hibernation tasks. To cover them, the five new tasks were sliced out of `hibernate.yml` (by name, from the file itself) and run twice with the role's own vars and handlers inside the CachyOS container, against a stub `/etc/default/limine` and a stub `limine-mkinitcpio`:

```
===== RUN 1 =====
TASK [Ensure ZRam is disabled in Limine cmdline]          changed
TASK [Ensure Zswap kernel params in Limine cmdline]       changed
TASK [Override CachyOS udev rule that disables Zswap]     changed
TASK [Ensure mkinitcpio drop-in directory exists]         changed
TASK [Add lz4 to initramfs MODULES]                       changed
RUNNING HANDLER [Rebuild initramfs]                       changed
localhost : ok=6 changed=6 unreachable=0 failed=0

===== RUN 2 (idempotency) =====
localhost : ok=5 changed=0 unreachable=0 failed=0

===== RESULTING STATE =====
# /etc/default/limine
KERNEL_CMDLINE[default]="quiet splash"
KERNEL_CMDLINE[default]+=" systemd.zram=0"
KERNEL_CMDLINE[default]+=" zswap.enabled=1 zswap.shrinker_enabled=1 zswap.compressor=lz4 zswap.max_pool_percent=30"

# /etc/udev/rules.d/30-zram.rules   (rule-free, shadows the cachyos-settings rule)
# /etc/mkinitcpio.conf.d/10-zswap.conf
MODULES+=(lz4)
```

The `mkinitcpio.conf.d` drop-in is equivalent to editing the `MODULES` array in `/etc/mkinitcpio.conf`: `/usr/bin/mkinitcpio` (41.1-2) concatenates the main config with `/etc/mkinitcpio.conf.d/*.conf` into one file and sources it once, so `MODULES+=(lz4)` appends to the array. Confirmed in the container:

```
$ cat /etc/mkinitcpio.conf /etc/mkinitcpio.conf.d/10-zswap.conf > /tmp/merged.conf
$ . /tmp/merged.conf; declare -p MODULES
declare -a MODULES=([0]="lz4")
```

The drop-in was chosen over an in-place regex rewrite of `/etc/mkinitcpio.conf` because it is idempotent and non-destructive; the resulting `MODULES` array is identical.

Final confirmation that hibernate/resume works end to end requires a reboot on the GZ302 — it cannot be exercised in a container.

palazzem:
**Claude Harness**

Review-panel angle selection for this PR (deterministic rubric, one reviewer per angle):

**Core (always on)**
- `correctness` — selected
- `divergence` — selected (judged against issue #290 and its **Claude Harness** completeness/scope comment; fast-lane run, no separate design doc)
- `simplicity` — selected
- `settling` — selected

**Conditional**
- `security` — skipped: no authn/authz, secret handling, crypto, or input parsing/deserialization in the diff
- `reliability` — **fired**: the diff adds a new subprocess call across a process boundary — the `Rebuild initramfs` handler invoking `limine-mkinitcpio` (`roles/system/handlers/main.yml`), notified by three tasks in `roles/system/tasks/hibernate.yml`
- `performance` — skipped: no code added inside a loop body and no caching logic introduced (Zswap is kernel configuration, not a cache implementation in this codebase)

**Repo-supplied**
- none: no `.claude/review-angles/` directory in the checkout

roles/system/tasks/hibernate.yml
```diff
@@ -30,6 +72,30 @@
   changed_when: false
   become: true
 
+# zswap.compressor=lz4 is applied at boot, before the root filesystem is
+# mounted, so lz4 must ship inside the initramfs. mkinitcpio (>= 38)
+# concatenates /etc/mkinitcpio.conf.d/*.conf after the main config and
+# sources the result, so `MODULES+=(lz4)` appends to the MODULES array
+# without rewriting /etc/mkinitcpio.conf in place.
+- name: Ensure mkinitcpio drop-in directory exists
```
palazzem: **Claude Harness**

**`simplicity/F1` — This directory pre-create task is defensive generality: the mkinitcpio package already ships `/etc/mkinitcpio.conf.d/`, and no host that reaches this task can lack it**

- **Evidence:** `docker run --rm cachyos/cachyos:latest sh -c 'pacman -Sy --noconfirm mkinitcpio >/dev/null 2>&1; pacman -Ql mkinitcpio | grep conf.d; ls -ld /etc/mkinitcpio.conf.d'` → `mkinitcpio /etc/mkinitcpio.conf.d/` and `drwxr-xr-x ... /etc/mkinitcpio.conf.d` — the package itself owns and creates the directory. Traced path: `hibernate.yml` runs only `when: host_has_systemd` (`roles/system/tasks/main.yml:24`), and `playbook.yml` pre_tasks document that marker as "absent under `docker build` RUN" — so the container test (the one environment without mkinitcpio) never reaches this task. Every host that does reach it must have mkinitcpio installed for the `limine-mkinitcpio` call in the `Rebuild initramfs` handler this same file notifies at line 97; a host where the directory is missing is a host where the handler fails anyway, so the task adds zero coverage. The pre-existing systemd `.conf.d` pre-create at lines 103-111 exists because systemd genuinely does not ship those directories (its comment says so); that pattern was carried over here to a package that does ship its drop-in directory.
- **Impact:** A dead task on every run: one extra task in every provisioning pass and one extra block for readers to reason about, protecting against nothing. Deleting the `Ensure mkinitcpio drop-in directory exists` task (lines 80-85) changes no reachable behavior.

palazzem (commented):
**Claude Harness**

**Angle `simplicity` — 1 finding(s)**

Examined the ZRam-to-Zswap additions in `roles/system/tasks/hibernate.yml` (two Limine cmdline `lineinfile` tasks, the udev-rule shadow, the mkinitcpio drop-in pair), the new `Rebuild initramfs` handler in `roles/system/handlers/main.yml`, and the two kernel-param vars in `roles/system/vars/main.yml`, looking for abstraction, indirection, or generality no requirement demands. The var + pinned-regexp pattern mirrors the existing `roles/hardware/tasks/gz302.yml` convention and the role-local handler is deliberate and justified in its comment — neither is a finding. One task guards a state no supported target can exhibit.

- `simplicity/F1` — Pre-creating `/etc/mkinitcpio.conf.d` guards a state that cannot occur on any host that reaches the task → roles/system/tasks/hibernate.yml:80

palazzem (commented):
**Claude Harness**

Examined the diff (`roles/system/tasks/hibernate.yml`, `roles/system/handlers/main.yml`, `roles/system/vars/main.yml`) against the spec record: issue #290 body plus its **Claude Harness** completeness/scope comment (full CachyOS wiki "Switching from ZRam to Zswap" procedure, user-confirmed; no approved design doc — declared absent). Compared each of the five user-confirmed missing steps against the implementation, and re-fetched the wiki procedure to check for steps outside the enumeration. Step 1 (`systemd.zram=0`) and step 2 (`zswap.enabled=1 zswap.shrinker_enabled=1 zswap.compressor=lz4 zswap.max_pool_percent=30`) land verbatim as `KERNEL_CMDLINE[default]+=` lines in `/etc/default/limine`, matching the repo's established Limine pattern (`roles/hardware/tasks/gz302.yml`). Step 3's override is comment-only rather than empty — udev ignores comment lines, so it is functionally the empty file the wiki prescribes at the same path `/etc/udev/rules.d/30-zram.rules`. Step 4 uses an `/etc/mkinitcpio.conf.d/10-zswap.conf` drop-in with `MODULES+=(lz4)` instead of editing `/etc/mkinitcpio.conf` in place; mkinitcpio.conf(5) confirms drop-in snippets in `/etc/mkinitcpio.conf.d/*.conf` are read after the main config, so the append reaches the MODULES array. Step 5's rebuild is the `Rebuild initramfs` handler running `limine-mkinitcpio`, the exact Limine command the wiki and the spec's environment facts name. The pre-existing Btrfs subvolume/swapfile/fstab/swapon steps the spec marks as already covered are unchanged. No departure, omission, or invention against the spec record was found.

roles/system/tasks/hibernate.yml
```diff
@@ -1,4 +1,46 @@
...
+# Kernel params live in /etc/default/limine; limine-mkinitcpio regenerates
+# the boot entries and the initramfs from it, so both param tasks and the
+# mkinitcpio drop-in below share the same handler.
+- name: Ensure ZRam is disabled in Limine cmdline
+  ansible.builtin.lineinfile:
+    path: /etc/default/limine
+    regexp: '^KERNEL_CMDLINE\[default\]\+=" systemd\.zram='
+    line: 'KERNEL_CMDLINE[default]+=" {{ system_zram_kernel_param }}"'
+  become: true
+  notify: Rebuild initramfs
```
palazzem: **Claude Harness**

**`reliability/F1` — A failure in any task after this point permanently loses the `Rebuild initramfs` notification: `/etc/default/limine` diverges from the boot entries/initramfs and no rerun ever converges them**

- **Evidence:** Handlers flush at end of play, and `playbook.yml:83-131` imports ten more roles (`desktop`, `virtualization`, `containers`, `trust`, `languages`, `tools`, `ai`, `security`, `home` — including network-dependent AUR builds) into the same play after `system`. Neither `ansible.cfg` nor `playbook.yml` sets `force_handlers` (`grep -rn force_handlers` over the repo returns nothing). Ansible docs, *Error handling in playbooks → Handlers and failure*: "By default, handlers do not run if a task fails later in the play, which can leave systems in an inconsistent state." Reproduced in an isolated `hanzo:test` container with a minimal playbook modeling this structure (config task notifies handler, later task fails): run 1 → `ok=1 changed=1 failed=1`, config written, handler marker absent; run 2 (failure fixed) → `ok=2 changed=0`, config task reports `ok`, **handler still never runs** — the rebuild marker does not exist after any number of reruns.
- **Impact:** `/etc/default/limine` contains `systemd.zram=0` + the `zswap.*` params and `/etc/mkinitcpio.conf.d/10-zswap.conf` exists, but `limine-mkinitcpio` never runs: the machine keeps booting with ZRam active and no lz4 in the initramfs, so the exact bug #290 fixes (hibernation never resumes) persists — while every subsequent `hanzo` run reports fully clean, hiding the divergence. `force_handlers: true` on the play (or `meta: flush_handlers` after the notifying tasks) closes this window.

roles/system/handlers/main.yml
```diff
@@ -1,6 +1,16 @@
 ---
 # logind reads its config on SIGHUP; a full restart would terminate
 # every active login session. Supported on systemd v243+.
+# Deliberately role-local rather than reusing the hardware role's
+# equivalent handler: handler names resolve globally, and sharing one
+# would tie the system role's hibernation tasks to a role that only runs
+# on matching hardware. limine-mkinitcpio regenerates the initramfs and
+# rewrites the Limine boot entries from /etc/default/limine.
+- name: Rebuild initramfs
+  ansible.builtin.command: limine-mkinitcpio
```
palazzem: **Claude Harness**

**`reliability/F2` — A single failed `limine-mkinitcpio` invocation is never retried: the notification is consumed by the failing handler run, and every rerun reports clean while boot entries and initramfs stay stale**

- **Evidence:** Reproduced in an isolated `hanzo:test` container with a minimal playbook modeling this handler (config task notifies; handler command exits non-zero): run 1 → handler runs and fails (`ok=2 changed=1 failed=1`), config already written; run 2 with the handler command healthy again → `ok=2 changed=0`, no task changes, no notification, **handler never re-runs** — the rebuild marker never appears. Note `force_handlers` does not cover this case: the handler *did* run and failed, so recovery requires re-notification that idempotent `lineinfile`/`copy` tasks will never produce. `limine-mkinitcpio` failing once is realistic — transient ESP-full conditions, a mid-transaction pacman kernel state, or an interrupted run — and this handler has no `retries` and no post-condition check.
- **Impact:** After one transient subprocess failure, the system is stuck in the same divergent state as F1 — ZRam active, no lz4 in initramfs, hibernation broken — with provisioning permanently reporting clean; only a manual `limine-mkinitcpio` run recovers it. A convergence guard is needed in addition to any `force_handlers` fix — e.g. register success in a marker consulted by a verification task, or verify the generated Limine entries actually contain `systemd.zram=0` and re-notify when they don't.

palazzem (commented):
**Claude Harness**

**Angle `reliability` — 2 finding(s)**

Examined the new `Rebuild initramfs` handler (`roles/system/handlers/main.yml`) invoking `limine-mkinitcpio` as a subprocess and the three tasks notifying it in `roles/system/tasks/hibernate.yml`, in the context of `playbook.yml`'s single-play structure (10 roles imported after `system`) and `ansible.cfg` (no `force_handlers`). Verified failure/rerun semantics by reproducing them with a minimal playbook inside the `hanzo:test` container and against current Ansible documentation.

- `reliability/F1 — A failure in any later task permanently loses the initramfs rebuild; reruns never converge → roles/system/tasks/hibernate.yml:18`
- `reliability/F2 — A failed limine-mkinitcpio run is never retried; subsequent runs report clean while boot artifacts stay stale → roles/system/handlers/main.yml:10`

roles/system/handlers/main.yml
```diff
@@ -1,6 +1,16 @@
 ---
 # logind reads its config on SIGHUP; a full restart would terminate
 # every active login session. Supported on systemd v243+.
+# Deliberately role-local rather than reusing the hardware role's
+# equivalent handler: handler names resolve globally, and sharing one
+# would tie the system role's hibernation tasks to a role that only runs
+# on matching hardware. limine-mkinitcpio regenerates the initramfs and
+# rewrites the Limine boot entries from /etc/default/limine.
+- name: Rebuild initramfs
```
palazzem: **Claude Harness**

**`settling/F1` — Cloning `limine-mkinitcpio` into a second, differently-named handler extends the role-local-handler convention past where it holds: Ansible dedupes handler runs per name, so a first GZ302 provision rebuilds every initramfs twice**

- **Evidence:** Traced path. `playbook.yml` imports `hardware` and `system` in one play. On a first GZ302 provision, `roles/hardware/tasks/gz302.yml` changes `amdgpu.conf` (template) and the OLED cmdline (`lineinfile`) — both `notify: Regenerate initramfs` (`roles/hardware/handlers/main.yml`: `ansible.builtin.command: limine-mkinitcpio`) — while this PR's `hibernate.yml` changes the two Limine cmdline lines and the mkinitcpio drop-in — all `notify: Rebuild initramfs` (this handler: the identical `limine-mkinitcpio`). Ansible docs (playbooks_handlers.rst, devel): "If the same handler is notified multiple times, it will only run once" — dedupe is per handler *name*, so at play-end flush both handlers fire and `limine-mkinitcpio` executes twice against identical `/etc/default/limine` and mkinitcpio config; the first rebuild's output is fully discarded. The comment above justifies the clone — "handler names resolve globally, and sharing one would tie the system role's hibernation tasks to a role that only runs on matching hardware" — but the same docs page shows that dichotomy is false: "There is only one global, play-level scope for handlers regardless of where the handlers are defined, either in the `handlers:` section or in roles." A single `Rebuild initramfs` handler in `playbook.yml`'s play-level `handlers:` section is notifiable from both roles, couples neither role to the other, and restores per-name dedupe — the same playbook-level home this repo already uses for cross-role shared state (CLAUDE.md rule 3: shared detection facts live in `playbook.yml` `pre_tasks`).
- **Impact:** Every provision run in which both roles touch boot config — guaranteed on the first run on the target GZ302 hardware, and recurring whenever a hardware kernel param and a hibernation value change together — pays for two back-to-back full `limine-mkinitcpio` runs (initramfs generation for every installed kernel plus Limine entry rewrite), doubling the most expensive step of the play for zero effect. Each future role needing an initramfs rebuild inherits the precedent and adds another clone, multiplying the rebuilds further.

palazzem (commented):
**Claude Harness**

**Angle `settling` — 1 finding(s)**

Examined every convention this diff extends rather than invents: the `KERNEL_CMDLINE[default]+=" …"` per-param `lineinfile` pattern inherited from `roles/hardware/tasks/gz302.yml` (sound — the file is bash-sourced array-append syntax with no drop-in dir, and the prefix-anchored regexps keep both new tasks idempotent and updatable), the `# Managed by Hanzo` copy-with-content drop-in pattern, the directory pre-create pattern from a102d58, the `vars/main.yml` placement of the new kernel-param values, and the role-local `ansible.builtin.command` handler pattern cloned from `roles/hardware/handlers/main.yml`. The last one is the finding: the clone crosses from convention into defect because Ansible deduplicates handler runs per name.

- `settling/F1 — Cloning the hardware role's limine-mkinitcpio handler under a second name defeats Ansible's per-name handler dedupe: a first GZ302 provision runs limine-mkinitcpio twice → roles/system/handlers/main.yml:9`

roles/system/tasks/hibernate.yml
```diff
@@ -1,4 +1,46 @@
...
+# at priority 100 and cachyos-settings ships a udev rule that force-disables
+# Zswap on every zram0 change event. RAM-backed swap cannot hold a
+# hibernation image, so the machine never resumes. The tasks below follow
+# the CachyOS wiki "Switching from ZRam to Zswap" procedure:
+# https://wiki.cachyos.org/configuration/general_system_tweaks/#swap-configuration
+
+# Kernel params live in /etc/default/limine; limine-mkinitcpio regenerates
+# the boot entries and the initramfs from it, so both param tasks and the
+# mkinitcpio drop-in below share the same handler.
+- name: Ensure ZRam is disabled in Limine cmdline
```
palazzem, line 12: **Claude Harness**

**`testing-quality/F1` — No test executes any of this PR's tasks: every documented test path skips the `Configure hibernation` include, so the green `test` check cannot fail regardless of what these tasks contain**

- **Evidence:** `roles/system/tasks/main.yml:23-25` gates this file on `when: host_has_systemd`, derived in `playbook.yml` from `stat /run/systemd/system` — a path the playbook's own comment notes is "absent under `docker build` RUN". The CI run that gated this PR confirms: `gh run view 32062009503 --job 95485192663 --log | grep -A1 "Configure hibernation"` → `TASK [system : Configure hibernation]` / `skipping: [localhost]`, and none of the six new task names appear anywhere in that log. Reproduced in an isolated worktree at head `1bd1747`: `docker build -f tests/Containerfile .` exits 0 with recap `failed=0`, and `ansible-playbook playbook.yml --check --diff --tags system` inside the built image prints the same skip. The gate is independent of `ANSIBLE_ARGS`, so all four test commands documented in CLAUDE.md — including the real-run `ANSIBLE_ARGS=""` — skip identically. (`/etc/default/limine` also does not exist in `cachyos/cachyos:latest`: `docker run --rm cachyos/cachyos:latest ls /etc/default/limine` → `No such file or directory`, so nothing has ever executed the two `lineinfile` tasks.)
- **Impact:** Spec #290's claim — hibernation resumes once the wiki's ZRam→Zswap steps are applied — is asserted by nothing that can fail. A wrong `regexp`, a typo in `system_zswap_kernel_params`, a broken `limine-mkinitcpio` handler invocation, or a silently incomplete task set (the exact bug class this PR fixes) would ship behind the same green check that approved this PR.

palazzem (commented):
**Claude Harness**

Examined whether any test can fail if this PR's fix is wrong: the six new tasks and the `Rebuild initramfs` handler across `roles/system/tasks/hibernate.yml`, `roles/system/handlers/main.yml`, and `roles/system/vars/main.yml`, against the repo's only test surface (`tests/Containerfile` → `ansible-playbook`, driven by `.github/workflows/test.yaml`) and spec #290's claim that applying the CachyOS ZRam→Zswap steps makes hibernation work. Verified by rebuilding the test container from PR head `1bd1747` in an isolated worktree and by reading the log of the CI run that gated this PR.

- testing-quality/F1 — no test executes any task in this PR: every test path skips the `Configure hibernation` include, so the green `test` check proves nothing about the change → roles/system/tasks/hibernate.yml:12

palazzem (commented):
**Claude Harness**

Examined the three changed files (`roles/system/tasks/hibernate.yml`, `roles/system/handlers/main.yml`, `roles/system/vars/main.yml`) against issue #290 and the CachyOS wiki procedure the PR claims to implement. Verified: (1) all five spec steps are present with kernel params matching the wiki verbatim (`systemd.zram=0`; `zswap.enabled=1 zswap.shrinker_enabled=1 zswap.compressor=lz4 zswap.max_pool_percent=30`) and zram-generator docs confirm `systemd.zram=0` means "no zram devices will be created by the generator"; (2) both `lineinfile` tasks were executed in an isolated CachyOS container against a realistic `/etc/default/limine` — they append the exact lines, coexist with the pre-existing `amdgpu.dcdebugmask` line, and a second run reports `changed=0` (idempotent); (3) `cachyos-settings 1:1.4.0-1` ships `/usr/lib/udev/rules.d/30-zram.rules` that runs `echo N > /sys/module/zswap/parameters/enabled` on zram0 change, and udev(7) confirms same-named files in `/etc` replace it while `#` lines are ignored, so the comment-only override neutralizes it; (4) mkinitcpio 41.1 in the container defaults `_optconfd=1` and concatenates `/etc/mkinitcpio.conf.d/*.conf` after the main config before sourcing, so `MODULES+=(lz4)` appends as claimed, and the drop-in directory is pre-created before the copy; (5) the new `Rebuild initramfs` handler name does not collide with the hardware role's `Regenerate initramfs`; (6) the full container `--check` build passes with `failed=0`, with `hibernate.yml` correctly skipped under `docker build` via the `host_has_systemd` gate. No logic, boundary, or error-path defect found.

roles/system/tasks/hibernate.yml
palazzem: **Claude Harness**

**fix** — `fd2d932`. Confirmed the claim independently: `docker run --rm --platform=linux/amd64 cachyos/cachyos:latest sh -c 'pacman -Sy --noconfirm mkinitcpio; pacman -Ql mkinitcpio | grep conf.d; ls -ld /etc/mkinitcpio.conf.d'` →

```
mkinitcpio 41.1-2
mkinitcpio /etc/mkinitcpio.conf
mkinitcpio /etc/mkinitcpio.conf.d/
drwxr-xr-x 1 root root 0 Aug 14 17:21 /etc/mkinitcpio.conf.d
```

The package owns and creates the directory, and any host reaching this task must have mkinitcpio installed for the `limine-mkinitcpio` rebuild to work at all — so the pre-create protected nothing. Task deleted; the comment above the drop-in copy now records why this drop-in dir needs no pre-create while the systemd ones below still do.

Verified after the change: `pre-commit run --all-files` → all hooks pass; `docker build -f tests/Containerfile -t hanzo:test .` → exit 0, `PLAY RECAP ... ok=58 changed=30 failed=0 skipped=27`.

roles/system/tasks/hibernate.yml
palazzem: **Claude Harness**

**defer** — real, but repo-wide and out of this bug fix's bounds. Tracking issue: #293.

The finding is accurate and I did not contest it: no `force_handlers` anywhere (`grep -rn force_handlers` over the repo returns nothing), ten roles import after `system` in the same play, and idempotent `lineinfile`/`copy` tasks never re-notify on a later run.

Why it does not land here:

- The exposure is a property of every handler in this playbook, not of this change. `roles/hardware/handlers/main.yml` → `Regenerate initramfs` writes the *same* `/etc/default/limine`-derived boot artifacts and has the identical window; so do `Reload logind` and `Restart keyd`, both predating this PR (`Reload logind` is notified by the pre-existing `logind` drop-in task in this very file).
- The remedy is play-level. `force_handlers: true` on `playbook.yml` changes error-handling semantics for all eleven roles — including the AUR/network-dependent ones whose failures currently *stop* subsequent state changes — and `meta: flush_handlers` inside a role flushes every pending handler play-wide, reordering `hardware`'s rebuild too. That is a cross-cutting behavior decision, not an execution detail of #290, whose user-confirmed scope ruling enumerates exactly the five CachyOS wiki steps.

#293 records both the notification-loss window and the F2 no-retry gap with the option space, so the fix can be made once for all handlers rather than half-made here.

roles/system/handlers/main.yml
palazzem: **Claude Harness**

**defer** — real, but its remedy is a new convergence mechanism the spec record does not settle. Tracking issue: #293 (filed for this finding together with `reliability/F1`; both are the same divergence reached by different routes).

Accepted as stated: a handler that runs and fails consumes its notification, and no idempotent task will ever re-notify it — `force_handlers` does not help.

Why it does not land here:

- Identical exposure on the pre-existing `roles/hardware/handlers/main.yml` → `Regenerate initramfs`, which runs the same `limine-mkinitcpio` for the GZ302 kernel params. Fixing convergence for the hibernation handler only would leave the same boot artifacts stale by the other route.
- The fix you sketch — a success marker consulted by a verification task, or verifying the generated Limine entries contain the expected params and re-notifying — introduces persistent provisioning state and a verify-then-repair pattern this repo does not have. Choosing where that state lives and what the post-condition asserts is a design call outside the #290 scope ruling (the five wiki steps), so I am not making it inside a fast-lane bug fix.

#293 carries the reproduction semantics and both candidate directions.

roles/system/handlers/main.yml
palazzem: **Claude Harness**

**defer** — the double rebuild is real; the sound remedy restructures the `hardware` role and sets a handler-placement convention this repo has never used, so it is an owner call, not a bug-fix detail. Tracking issue: #294.

I reproduced both candidate remedies in the `hanzo:test` container with a two-role playbook modelling `playbook.yml` (per-role tags on `import_role`, both roles notifying an initramfs handler):

- **Play-level shared handler** (your proposal): full run → `RUNNING HANDLER [Regenerate initramfs]` exactly once; `--tags beta` (one role executing) → still runs once. Works, including under tag filtering.
- **Same name in both role-local handler files** (the cheap in-role variant): full run printed only `RUNNING HANDLER [beta : Regenerate initramfs]` — the first role's handler definition never ran, its notification silently resolved to the later definition. Dedupes by shadowing, correct only while both bodies stay byte-identical. Not a fix I would ship.

So the only sound option moves `Regenerate initramfs` out of `roles/hardware/handlers/main.yml` into `playbook.yml`'s `handlers:` section and rewires the hardware role's `notify:`. That edits a role outside this change's scope and establishes "cross-role handlers live in the play" as repo convention. Issue #290's user-confirmed scope ruling is the five CachyOS wiki steps; the impact here is duplicated work on runs where both roles change boot config, not a wrong end state (`limine-mkinitcpio` is idempotent — the second run regenerates identical output). #294 records the finding and the verified option space so the convention gets decided once, including for the next role that needs a rebuild.

roles/system/tasks/hibernate.yml
palazzem: **Claude Harness**

**defer** — accurate, and already the stated limitation of this run. Tracking issue: #295 (cross-referenced from the pre-existing #243).

Confirmed independently: `roles/system/tasks/main.yml` gates the include on `host_has_systemd`, `playbook.yml` derives it from `stat /run/systemd/system` whose own comment records it as "absent under `docker build` RUN", and the container build at head reports the hibernation include `skipping: [localhost]` with none of the new task names in the log. `/etc/default/limine` is likewise absent from `cachyos/cachyos:latest`.

Why it is not fixed in this PR: closing it means changing the test harness contract — either a systemd-capable surface (systemd as PID 1, or a VM job) so the gate evaluates true, or fixture-based execution that forces `host_has_systemd` and pre-seeds `/etc/default/limine` and the mkinitcpio drop-in dir, with content assertions and an idempotency pass. Both are new test infrastructure and a decision about what the repo's test surface guarantees; issue #290's scope ruling is the five CachyOS wiki steps. This is exactly the constraint the reproduction-evidence comment on this PR states up front, which is why the bug-lane evidence is recorded as commands + output against the same base image the test container uses rather than as a failing automated test.

#295 states the blocker concretely (the `host_has_systemd` gate, added in #282, postdates #243's general coverage discussion) and lists both directions.

roles/system/handlers/main.yml
palazzem: do not defer this refactor. Improve the code quality by following the suggestion.

palazzem: **Claude Harness**

**fix** — `ed5529e`. Defer withdrawn per your instruction; implemented the sound remedy you proposed.

`Regenerate initramfs` now lives in `playbook.yml`'s play-level `handlers:` section. `roles/hardware/handlers/main.yml` and `roles/system/handlers/main.yml` no longer define it (each keeps a comment pointing at the play-level home); the system role's three hibernation `notify:` lines were renamed from `Rebuild initramfs` to `Regenerate initramfs`, and the hardware role's two GZ302 `notify:` lines are unchanged. `limine-mkinitcpio` is therefore defined once and deduped once, and the next role needing a rebuild notifies the same handler instead of adding a clone.

Verified in the `hanzo:test` container with a two-role playbook modelling `playbook.yml` (per-role tags on `import_role`, both roles notifying the play-level handler, handler appending to a log):

```
### FULL RUN (both roles change boot config)
RUNNING HANDLER [Regenerate initramfs] *****
localhost : ok=3 changed=3 failed=0
--- handler invocations: 1

### TAG-FILTERED RUN (--tags beta only)
RUNNING HANDLER [Regenerate initramfs] *****
localhost : ok=2 changed=2 failed=0
--- handler invocations: 1
```

One run each, and selective runs still fire it. I also checked whether the handler needs a tag to survive `--tags` filtering: with the tag removed the `--tags beta` run still printed `RUNNING HANDLER [Regenerate initramfs]` with 1 invocation — notified handlers are not tag-filtered — so the handler is left untagged and a comment records why.

Verified after the change: `pre-commit run --all-files` → all hooks pass (ansible-lint included); `docker build -f tests/Containerfile -t hanzo:test .` → exit 0, `PLAY RECAP ... ok=58 changed=30 failed=0 skipped=27` — unchanged from the previous head.

Tracking issue #294 is now redundant for this finding; closing it as fixed here.

roles/hardware/handlers/main.yml
```diff
@@ -9,10 +9,9 @@
   changed_when: true
   become: true
 
-- name: Regenerate initramfs
-  ansible.builtin.command: limine-mkinitcpio
-  changed_when: true
-  become: true
+# `Regenerate initramfs`, notified by the GZ302 tasks in this role, is
```
palazzem: Remove this comment, unneeded.

roles/system/handlers/main.yml
```diff
@@ -1,4 +1,8 @@
 ---
+# `Regenerate initramfs`, notified by the hibernation tasks in this role,
```
palazzem: Remove this comment, unneeded

playbook.yml
```diff
@@ -137,3 +137,20 @@
             state: absent
           become: true
           tags: [always]
+
+  # Cross-role handlers live in the play. Handler scope is play-wide
```
palazzem: Simplify the comment as it just needs to mention what this handler does.

roles/hardware/handlers/main.yml
palazzem: **Claude Harness**

**fix** — `2fd9f26`. Comment removed from `roles/hardware/handlers/main.yml`; the file now goes straight from `Restart network stack` to `Restart keyd`.

Verified after the change: `pre-commit run --all-files` → all hooks pass (ansible-lint included); `docker build -f tests/Containerfile -t hanzo:test .` → exit 0, `PLAY RECAP ... ok=58 changed=30 failed=0 skipped=27` — unchanged from the previous head.

roles/system/handlers/main.yml
palazzem: **Claude Harness**

**fix** — `2fd9f26`. Comment removed from `roles/system/handlers/main.yml`; the file now opens directly on the `Reload logind` handler and its SIGHUP rationale.

Verified after the change: `pre-commit run --all-files` → all hooks pass (ansible-lint included); `docker build -f tests/Containerfile -t hanzo:test .` → exit 0, `PLAY RECAP ... ok=58 changed=30 failed=0 skipped=27` — unchanged from the previous head.

playbook.yml
palazzem: **Claude Harness**

**fix** — `2fd9f26`. The eleven-line comment above `handlers:` in `playbook.yml` is now two lines stating only what the handler does:

```yaml
  # Regenerates the initramfs for every installed kernel and rewrites the
  # Limine boot entries from /etc/default/limine.
  handlers:
    - name: Regenerate initramfs
      ansible.builtin.command: limine-mkinitcpio
```

The dedupe/placement and tag-filtering rationale is dropped from the source; it stays on the record in this PR (thread `settling/F1`) rather than in the playbook.

Verified after the change: `pre-commit run --all-files` → all hooks pass (ansible-lint included); `docker build -f tests/Containerfile -t hanzo:test .` → exit 0, `PLAY RECAP ... ok=58 changed=30 failed=0 skipped=27` — unchanged from the previous head.

Issue #292, opened 2026-08-17
Replace ollama with llama.cpp for local AI inference


palazzem:
## Problem statement

The machine's local AI inference stack is provisioned as ollama (the ROCm variant). Ollama wraps llama.cpp behind its own daemon and model-registry machinery, which hides the inference controls that matter to the owner: quantization choices, GPU offload tuning, and server flags are abstracted away or exposed only partially. The owner wants direct control over how models are run on their AMD GPU, and the wrapper layer costs them that control while adding nothing they use — no other tool on the machine depends on ollama's API or CLI. GPU-offloaded inference demonstrably works on this machine today under ollama-rocm, so a working ROCm path for this GPU exists.

## Expected behavior

Provisioning installs llama.cpp instead of ollama. After a provisioning run, the owner can run GPU-accelerated inference on their AMD GPU directly with llama.cpp's tools, and can start llama.cpp's server by hand whenever they want a local API — nothing runs in the background until they start it. The ollama entry is removed from provisioning entirely; anything ollama already left on the machine is the owner's own, separate cleanup task.

## Constraints

- Replace ollama with llama.cpp — the tool choice is the owner's directive; no other inference stack is in scope.
- Use the CachyOS-provided llama.cpp packaging — the owner's directive; upstream Arch/AUR packaging churn is irrelevant.
- The ROCm/HIP backend is preferred, matching today's ollama-rocm path.
- Inference must be GPU-accelerated on the machine's AMD GPU; CPU-only llama.cpp does not satisfy the goal.
- No always-on service: provisioning must not enable or start any inference server at boot or in the background.
- No removal or cleanup of existing ollama installs — provisioning stops installing it, nothing more. The owner explicitly owns manual cleanup of anything ollama left behind, as a separate task outside this change.
- Model files are the owner's business: provisioning installs tooling only and manages no models.

## Goals

1. **G-1** — After provisioning, llama.cpp's inference tooling (command-line inference and its server) is installed and usable by the provisioned user.
2. **G-2** — Inference with llama.cpp runs on the AMD GPU, with the provisioned user having the device access that requires.
3. **G-3** — The llama.cpp server is start-on-demand only: available to launch manually, never enabled as a background service by provisioning.
4. **G-4** — Provisioning no longer installs ollama: the ollama entry is removed from the provisioner's configuration.

## Non-goals

- Removing or cleaning up ollama (package, models, state, any hand-enabled service) from already-provisioned machines — explicitly the owner's own separate task.
[...]

PR #296, opened 2026-08-17
feat(ai)!: replace ollama with llama.cpp on the ROCm backend

palazzem:
**Claude Harness**

Replaces ollama with llama.cpp as the machine's local inference stack, per #292: the `ai` role now installs the CachyOS-provided `llama-cpp` package (CLI + `llama-server`) together with the `ggml-hip` (ROCm/HIP) and `ggml-cpu` backends, which upstream ships as separate, dynamically loaded packages. The `render`/`video` group membership that AMD GPU compute needs is unchanged, nothing is enabled at boot — `llama-cpp` ships `llama-server.service` units that the role deliberately leaves disabled — and no ollama package reference remains in provisioning. Existing ollama installs are untouched; their cleanup is the owner's separate task.

## Run record

- **Issue:** #292
- **Session:** `2b073410-e8fc-4c6e-80d5-a52268b29770` — resume with `claude --resume 2b073410-e8fc-4c6e-80d5-a52268b29770`

### DIFF
--- a/roles/ai/tasks/main.yml
+++ b/roles/ai/tasks/main.yml
@@ -10,10 +10,14 @@
     name: "{{ ai_pacman }}"
   become: true
 
+# No check-mode gate: render and video are static system groups (systemd
+# sysusers.d/basic.conf), so they exist in the test container too, and the
+# user module resolves every name in `groups` even under --check. Running
+# the task in the container build is what makes a missing or renamed group
+# fail the run instead of passing green on a skip.
 - name: Add user to AI supplementary groups
   ansible.builtin.user:
     name: "{{ ansible_facts['user_id'] }}"
     groups: "{{ ai_user_groups }}"
     append: true
-  when: not ansible_check_mode
   become: true
--- a/roles/ai/tasks/main.yml
+++ b/roles/ai/tasks/main.yml
@@ -1,4 +1,10 @@
 ---
+# Local inference is start-on-demand: the llama-cpp package ships
+# llama-server.service (system and user units), and this role
+# deliberately never enables or starts either. Nothing listens until
+# the owner launches llama-server (or a llama-cli run) by hand.
+#
+# Model files are out of scope — this role installs tooling only.
 - name: Install AI pacman packages
   community.general.pacman:
     name: "{{ ai_pacman }}"
--- a/roles/ai/vars/main.yml
+++ b/roles/ai/vars/main.yml
@@ -22,11 +22,11 @@ ai_pacman:
 # ROCm requires membership in both render and video per AMD docs
 # (https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/prerequisites.html):
 # render gates DRM render-node access, video gates the legacy
-# video-device path. Both are kernel-managed (udev creates them when
-# GPU devices appear), so they're absent in containers — gating
-# consumed by the role's user task. video is also declared in the
-# desktop role's desktop_user_groups for general graphics access;
-# the duplicate append is idempotent.
+# video-device path. Both are static system groups declared in systemd's
+# sysusers.d/basic.conf, so they exist on any systemd host and in the
+# test container — the role's user task therefore runs unconditionally.
+# video is also declared in the desktop role's desktop_user_groups for
+# general graphics access; the duplicate append is idempotent.
 ai_user_groups:
   - render
   - video
--- a/roles/ai/vars/main.yml
+++ b/roles/ai/vars/main.yml
@@ -1,6 +1,23 @@
 ---
+# llama.cpp from the CachyOS-provided repos (Arch `extra`, mirrored by
+# CachyOS). Upstream splits the runtime into the frontend binaries
+# (llama-cpp) and one package per ggml backend, loaded dynamically at
+# runtime — so the backend set is chosen here, by which ggml-* packages
+# get installed.
 ai_pacman:
-  - ollama-rocm
+  # CLI tooling and the HTTP server: /usr/bin/llama-cli,
+  # /usr/bin/llama-server, plus llama-bench/llama-quantize et al.
+  - llama-cpp
+  # ROCm/HIP backend, the GPU path on this machine's AMD GPU; pulls
+  # rocm-hip-runtime, hipblas and rocblas as hard dependencies. Declared
+  # an optional dep by `ggml`, so it is not implied by llama-cpp and has
+  # to be requested explicitly.
+  - ggml-hip
+  # CPU backend, also an optional dep of `ggml`. Required for everything
+  # that stays on the host — tokenization and any layer not offloaded —
+  # so llama.cpp stays usable when a model exceeds VRAM or the GPU
+  # backend is unavailable.
+  - ggml-cpu
 
 # ROCm requires membership in both render and video per AMD docs
 # (https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/prerequisites.html):


palazzem:
**Claude Harness**

Verification evidence for the container-provable acceptance criteria (AC-1, AC-2, AC-4, AC-5) and EC-1. AC-3 and AC-6 are the owner's post-acceptance smoke tests on the real machine and are not run here.

**EC-1 — provisioning completes without an AMD GPU and without models**

`docker build -f tests/Containerfile -t hanzo:test .` (default `--check --diff`):

```
localhost                  : ok=58   changed=30   unreachable=0    failed=0    skipped=27   rescued=0    ignored=0
```

Real (non-check) run of the role, `docker build --build-arg ANSIBLE_ARGS="--tags ai" -f tests/Containerfile -t hanzo:test-ai .`:

```
TASK [ai : Install AI pacman packages] *****************************************
TASK [ai : Add user to AI supplementary groups] ********************************
localhost                  : ok=21   changed=10   unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
```

**AC-1 / AC-2 / AC-4 / AC-5 — inspected inside that provisioned image, as the provisioned user**

```
=== AC-1 binaries as provisioned user (testuser):
testuser
/usr/sbin/llama-cli
/usr/sbin/llama-server
0.00.288.294 E ggml_cuda_init: failed to initialize ROCm: no ROCm-capable device is detected
version: 0.1.0-dev (build 10453, commit 3cb7ffb1a1)
built with GNU 16.2.1 for Linux x86_64
0.00.114.677 E ggml_cuda_init: failed to initialize ROCm: no ROCm-capable device is detected
version: 0.1.0-dev (build 10453, commit 3cb7ffb1a1)
built with GNU 16.2.1 for Linux x86_64
=== AC-2 ollama packages installed:
grep rc=1 (1 = none installed)
=== AC-2 packages installed by the ai role:
llama-cpp b10453-1
ggml 0.20.0-1
ggml-hip 0.20.0-1
ggml-cpu 0.20.0-1
=== AC-4 group membership:
uid=1000(testuser) gid=1000(testuser) groups=1000(testuser),983(video),987(render),998(wheel)
=== AC-5 service units present, enablement state:
/usr/lib/systemd/system/llama-server.service
/usr/lib/systemd/user/llama-server.service
enabled-symlink grep rc=1 (1 = not enabled)
any llama/ollama unit link rc=1 (1 = none)
pgrep rc=1 (1 = nothing running)
```

Notes:

- The `ggml_cuda_init: failed to initialize ROCm: no ROCm-capable device is detected` line is the HIP backend being loaded and finding no GPU in the container — evidence that `ggml-hip` is the active backend package, and that the tooling degrades gracefully where no AMD GPU exists (EC-1). Actual layer offload is AC-3, on the owner's machine.
- AC-2 also holds at the source level: `grep -rni ollama` across the repo (excluding `.git`) returns no matches.
- EC-2 needs no code: the role only installs; nothing in it removes or touches an existing ollama install.

**Packaging note.** CachyOS' repos carry upstream's split packaging: `llama-cpp` (frontend binaries, `Replaces: llama-cpp-rocm llama-cpp-vulkan`) depends on `ggml`, which declares each backend as an optional dep loaded dynamically at runtime. So the ROCm/HIP path requires requesting `ggml-hip` explicitly, and `ggml-cpu` covers the host-side ops (tokenization, non-offloaded layers).

**Lint** — `pre-commit run --all-files`: all hooks pass (`ansible-lint ... Passed`).

palazzem:
**Claude Harness**

Review panel — angle selection for this round (deterministic against the diff: `roles/ai/vars/main.yml`, `roles/ai/tasks/main.yml`):

Core (always on):
- `correctness` — selected
- `divergence` — selected (spec issue #292; no design doc — fast lane)
- `simplicity` — selected
- `settling` — selected

Conditional:
- `security` — skipped (diff touches no authn/authz, secret handling, crypto, or input parsing)
- `data` — skipped (no schema, migration, or query text)
- `concurrency` — skipped (no parallel execution, shared state, locks, or async coordination)
- `reliability` — skipped (no new/changed cross-process call or retry/timeout/recovery logic; diff is a pacman package-list change and comments)
- `performance` — skipped (no loop-body code or caching)

Repo-supplied:
- none (no `.claude/review-angles/` in the checkout)

palazzem (commented):
**Claude Harness**

Examined the full diff — `roles/ai/vars/main.yml` (the `ai_pacman` swap from `ollama-rocm` to `llama-cpp` + `ggml-hip` + `ggml-cpu`, and the `ai_user_groups` list left as-is) and `roles/ai/tasks/main.yml` (comment-only change) — for abstraction, indirection, or generality no requirement demands. The one candidate was `ggml-cpu` as possibly undemanded generality or a redundant transitive dep; verified in a CachyOS container that `ggml-hip` does not depend on `ggml-cpu` and `ggml` lists it only as an optional dep, so the explicit entry is load-bearing for the dynamically-loaded-backend packaging, not speculative breadth. The `ai_pacman`/`ai_user_groups` variable indirection predates this PR and is the repo-wide role convention. No new abstraction, wrapper, or configurable surface is introduced.

palazzem (commented):
**Claude Harness**

**Angle `settling` — run and clean**

Examined the full diff — `roles/ai/vars/main.yml` (the `ai_pacman` swap from `ollama-rocm` to `llama-cpp` + `ggml-hip` + `ggml-cpu` and its comments) and `roles/ai/tasks/main.yml` (the start-on-demand comment block) — against every pre-existing convention the change extends. Checked: (1) the unconditional install of the hardware-specific ROCm backend, extending the old `ollama-rocm` pattern with no GPU detection fact — the playbook's detection facts gate execution failures, not package selection, and issue #292 mandates ROCm/HIP for this single-owner AMD machine while EC-1 only requires the container run to complete, which a package install does, so gating is not an evidently better approach; (2) the flat `vars/main.yml` package-list convention rather than overridable `defaults/` — endorsed by the project's CLAUDE.md and appropriate for non-overridable single-machine values; (3) the explicit listing of `ggml-cpu`/`ggml-hip` — verified against current Arch packaging (`ggml` 0.20.0 declares both only as optdepends), so the explicit listing is required, and the diff correctly avoided assuming `llama-cpp` alone pulls a backend. No extended convention is flawed with an evidently better alternative. No findings.

palazzem (commented):
**Claude Harness**

Examined the full diff (`roles/ai/vars/main.yml`, `roles/ai/tasks/main.yml`) against spec issue #292 (no approved design exists) for departures, omissions, and inventions. Verified in an isolated CachyOS container run: `llama-cpp`, `ggml-hip`, and `ggml-cpu` all resolve from the repo set CachyOS ships (`extra`; the `cachyos/*` repos carry no llama.cpp alternative), `llama-cli` and `llama-server` are installed and invocable by a non-root user (AC-1), `llama-server.service` (system and user units) is present but `disabled` with no preset entry and the role never enables it (AC-5/G-3), `ggml-hip` hard-depends on `rocm-hip-runtime`/`hipblas`/`rocblas` (ROCm constraint), `grep -rni ollama` over the tree returns nothing (AC-2/G-4), `render`/`video` group handling is intact (AC-4), the role adds no removal or model-management tasks (EC-2, non-goals), and the project's gating check-mode container build completes with `failed=0` (EC-1). No departure from the spec found.

palazzem (commented):
**Claude Harness**

Examined the full diff — `roles/ai/vars/main.yml` (the `ollama-rocm` → `llama-cpp`/`ggml-hip`/`ggml-cpu` swap and its claims) and `roles/ai/tasks/main.yml` (the start-on-demand comment) — and verified every factual claim against the live `cachyos/cachyos:latest` image: all three packages resolve from `extra` (`pacman -Si`); `ggml` declares both `ggml-cpu` and `ggml-hip` as *optional* deps, so listing them explicitly is required, as the comments state; `ggml-hip` hard-depends on `rocm-hip-runtime`, `hipblas`, `rocblas` as claimed; `llama-cpp` ships `/usr/bin/llama-cli`, `/usr/bin/llama-server`, `llama-bench`, `llama-quantize` plus both system and user `llama-server.service` units (`pacman -Fl`), and a real install leaves the service `disabled` with no preset enabling it (AC-5); `llama-cpp` neither conflicts with nor replaces `ollama-rocm` — both install side by side (EC-2) — and `llama-cli --version` runs after install (AC-1). `grep -r ollama` over the branch returns nothing (AC-2), and the project's container gate (`docker build --build-arg ANSIBLE_ARGS="--tags ai --check --diff"`) completes with `failed=0`, the group task correctly skipped in check mode. No logic, boundary, or error-path defect found.

roles/ai/tasks/main.yml
```diff
@@ -1,4 +1,10 @@
 ---
+# Local inference is start-on-demand: the llama-cpp package ships
+# llama-server.service (system and user units), and this role
+# deliberately never enables or starts either. Nothing listens until
+# the owner launches llama-server (or a llama-cli run) by hand.
+#
+# Model files are out of scope — this role installs tooling only.
```
palazzem, line 7: **Claude Harness**

**`testing-quality/F1` — AC-4 gates acceptance in the container run, but that run skips the group-membership task and nothing else asserts membership**

- **Evidence:** Issue #292: "AC-1, AC-2, AC-4, and AC-5 are provable in the project's container test run and gate acceptance of the change" and "**AC-4** (G-2) — After provisioning, the provisioned user belongs to the groups required for AMD GPU compute access, verifiable via group membership." The gating build of this branch (`docker build -f tests/Containerfile`, default `--check --diff`) logs `TASK [ai : Add user to AI supplementary groups] ... skipping: [localhost]` — the task is gated `when: not ansible_check_mode`, and no assert/`getent`/`id` step exists anywhere in `tests/Containerfile`, `playbook.yml`, or `roles/ai/`. The exemption premise is also false: the vars comment claims render/video are "absent in containers", but `docker run --rm cachyos/cachyos:latest getent group render video` returns `render:x:987:` / `video:x:983:` — both groups exist in the test container, so AC-4 is verifiable there.
- **Impact:** a regression that drops `render`/`video` from `ai_user_groups` or breaks the append passes the acceptance gate green; AC-4 — the spec's GPU-access guarantee that does survive into the container — is accepted on faith, not proof.

roles/ai/tasks/main.yml
```diff
@@ -1,4 +1,10 @@
 ---
+# Local inference is start-on-demand: the llama-cpp package ships
+# llama-server.service (system and user units), and this role
+# deliberately never enables or starts either. Nothing listens until
```
palazzem, line 4: **Claude Harness**

**`testing-quality/F2` — AC-5's "no service enabled or running" is carried by this comment alone; no check in the repo can fail if it stops being true**

- **Evidence:** Issue #292: "**AC-5** (G-3) — After a provisioning run, no llama.cpp service is enabled or running; the server binary is present and not started by anything but the owner" — declared container-provable and acceptance-gating. The harness's only test step is `RUN ansible-playbook playbook.yml $ANSIBLE_ARGS` (tests/Containerfile) with default `--check`, which installs nothing and observes no unit state; `grep -rn 'systemctl\|is-enabled' tests/ roles/ai/` matches nothing but prose. The risk this AC names is real: a real in-container install (`pacman -S --noconfirm llama-cpp ggml-cpu`) shows the package ships `/usr/lib/systemd/system/llama-server.service` **and** `/usr/lib/systemd/user/llama-server.service`, currently with no enablement symlink under `/etc/systemd/system` and no preset entry — i.e. AC-5 holds today purely by upstream packaging behavior that nothing in this repo verifies.
- **Impact:** if a future `llama-cpp` package or a CachyOS preset enables `llama-server`, provisioning starts violating G-3's no-always-on-service constraint while every gate stays green — the spec's start-on-demand guarantee has no falsifier.

palazzem (commented):
**Claude Harness**

**Angle `testing-quality` — 3 finding(s)**

Examined whether the tests prove issue #292's gated claims for this diff (`roles/ai/vars/main.yml`, `roles/ai/tasks/main.yml`): the spec declares AC-1, AC-2, AC-4, AC-5 "provable in the project's container test run" and acceptance-gating, where the project's only harness is `tests/Containerfile` ending in `ansible-playbook playbook.yml --check --diff`. Verified in an isolated clone of this branch: the gating build passes (exit 0, EC-1); a mutation build with a bogus package name **fails** the gate (`Failed to list package llama-cpp-definitely-nonexistent-xyz`, exit 1), so package identity is covered by a check that can fail — AC-2 holds (plus `grep -ri ollama` over the tree: zero hits). The findings are the three gated properties the harness cannot falsify: AC-4 (task skipped in check mode), AC-5 (no service-state check exists), AC-1's "invocable" (nothing installs or executes the binaries).

- `testing-quality/F1` — AC-4's group-membership task is skipped by the acceptance-gating run and asserted by nothing → roles/ai/tasks/main.yml:7
- `testing-quality/F2` — AC-5's "no service enabled or running" exists only as a comment; no check can fail if it stops being true → roles/ai/tasks/main.yml:4
- `testing-quality/F3` — AC-1's "installed and invocable" is proven only as check-mode transaction resolution; no test path installs or invokes the binaries → roles/ai/vars/main.yml:10

roles/ai/vars/main.yml
```diff
@@ -1,6 +1,23 @@
...
+# llama.cpp from the CachyOS-provided repos (Arch `extra`, mirrored by
+# CachyOS). Upstream splits the runtime into the frontend binaries
+# (llama-cpp) and one package per ggml backend, loaded dynamically at
+# runtime — so the backend set is chosen here, by which ggml-* packages
+# get installed.
 ai_pacman:
-  - ollama-rocm
+  # CLI tooling and the HTTP server: /usr/bin/llama-cli,
+  # /usr/bin/llama-server, plus llama-bench/llama-quantize et al.
+  - llama-cpp
```
palazzem, line 10: **Claude Harness**

**`testing-quality/F3` — AC-1's "installed and invocable by the provisioned user" is proven only as far as check-mode transaction resolution; no test path installs or invokes the binaries**

- **Evidence:** Issue #292: "**AC-1** (G-1) — After a provisioning run, llama.cpp's command-line inference tool and server binary are installed and invocable by the provisioned user" — acceptance-gating. The gating run's entire coverage of this line is the pacman check-mode diff (`+llama-cpp-b10453-1`, `+ggml-hip-0.20.0-1`, `+ggml-cpu-0.20.0-1`, …): it proves the names resolve in the repos (a mutation to a bogus name does fail the build — verified), but nothing installs the packages or executes anything; `grep -rn llama tests/Containerfile playbook.yml` finds no invocation step. Invocability had to be proven out-of-repo: real install in a cachyos container, then `llama-cli --version` → `version: 0.1.0-dev (build 10453, commit 3cb7ffb1a1)`.
- **Impact:** a package that resolves but ships broken or renamed binaries passes the acceptance gate — not hypothetical churn for this package, which already absorbed renames (`pacman -Si llama-cpp` → `Replaces: llama-cpp-rocm llama-cpp-vulkan`); AC-1's operative property (invocable tools) is untested by anything that can fail.

roles/ai/tasks/main.yml
palazzem: **Claude Harness**

**Fix** — commit `6559499`.

Your premise checks out, and the vars comment that justified the gate was wrong on the facts:

```
$ docker run --rm --platform=linux/amd64 cachyos/cachyos:latest getent group render video
render:x:987:
video:x:983:
$ docker run --rm --platform=linux/amd64 cachyos/cachyos:latest grep -rn "render\|video" /usr/lib/sysusers.d/
/usr/lib/sysusers.d/basic.conf:35:g render  -     -            -
/usr/lib/sysusers.d/basic.conf:40:g video   -     -            -
# owned by systemd 261.2-1
```

`render`/`video` are static system groups declared by systemd's `sysusers.d/basic.conf`, not udev-created ones — they exist on every systemd host and in the test container. `6559499` drops `when: not ansible_check_mode` from the group task and rewrites the comment to state what is actually true.

The task now runs in the gating build instead of skipping:

```
$ docker build --progress=plain -f tests/Containerfile -t hanzo:test .
...
TASK [ai : Add user to AI supplementary groups] ********************************
changed: [localhost]
...
localhost                  : ok=59   changed=31   unreachable=0   failed=0   skipped=26
```

And it now carries a falsifier — `ansible.builtin.user` resolves every name in `groups` even under `--check`, so a dropped-to-nonexistent or renamed group fails the build. Mutation test (`ai_user_groups` + `hanzo-bogus-group`, reverted):

```
$ docker build --progress=plain --build-arg ANSIBLE_ARGS="--tags ai --check --diff" -f tests/Containerfile -t hanzo:test-ai-mutant .
TASK [ai : Add user to AI supplementary groups] ********************************
[ERROR]: Task failed: Module failed: Group hanzo-bogus-group does not exist
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Group hanzo-bogus-group does not exist"}
EXIT=1
```

Scope note, so the residual isn't silently claimed as covered: this makes the AC-4 task *executed and falsifiable* in the gating run, but `--check` still installs nothing and modifies no account, so post-run membership itself (`id`/`getent` on the provisioned user) is not asserted — that needs a real-install verification stage, which is the harness gap tracked in #301 alongside `F2`/`F3`. A regression that removes a group from the list without replacing it with an invalid name is caught by review, not by the build.

Lint green: `pre-commit run --all-files` → all hooks Passed.

roles/ai/tasks/main.yml
palazzem: **Claude Harness**

**Defer** — tracked in #301.

The finding is accurate and I'm not contesting any of it: `grep -rn 'systemctl\|is-enabled' tests/ roles/` matches nothing but prose, the packaged `llama-server.service` (system and user) is real, and AC-5 holds today only because upstream ships it unenabled with no preset entry — nothing in this repo can fail if that stops being true.

It is deferred rather than fixed because the missing falsifier cannot be built inside this change's bounds. The gating surface is `RUN ansible-playbook playbook.yml --check --diff`, which installs nothing and therefore has no unit state to observe: any check of enablement requires a *real-install* stage in `tests/Containerfile` plus a CI job that runs it (`.github/workflows/test.yaml` runs only the default `--check` build today). That is a change to the harness contract — runtime, network cost, and what "the test" means — affecting every role, not the `ai` role; #292's constraints scope this change to the package swap and its group/no-service properties, and its acceptance-criteria preamble puts the harness itself out of scope. Folding a new verification stage into a package swap would be the design call I'm not entitled to make here.

What this PR does contribute to AC-5 is negative-only and stated as such: the role contains no `systemd`/`service`/`systemctl` task at all, so provisioning cannot enable or start anything (`roles/ai/tasks/main.yml` is two tasks — pacman install, user groups), and the header comment records the intent so a future "enable it for convenience" edit reads as a spec violation rather than a nicety.

#301 carries your evidence verbatim — the shipped unit paths, the absent enablement symlink and preset entry, and the risk that a future package revision or CachyOS preset flips it — and lists `systemctl is-enabled llama-server.service` in a real-install stage as the concrete falsifier to add. Resolving here; #301 is the open record.

roles/ai/vars/main.yml
palazzem: **Claude Harness**

**Defer** — tracked in #301.

Accurate as written: the gating build's coverage of AC-1 stops at pacman check-mode transaction resolution, and your out-of-repo proof (`llama-cli --version` → `version: 0.1.0-dev (build 10453, commit 3cb7ffb1a1)` after a real install) is exactly the step that has no in-repo home. The rename history you cite (`Replaces: llama-cpp-rocm llama-cpp-vulkan`) makes the risk concrete rather than theoretical.

Deferred for the same structural reason as `F2`: proving invocability requires *installing* the package, which the gating `--check` run never does. The only surface where `llama-cli --version` could run is the real-run build (`ANSIBLE_ARGS=""`), which is documented in `CLAUDE.md` but not asserted on and not run by CI. Adding it means a new CI stage and a convention for post-install assertions that would apply to every role in the repo — a harness-contract decision outside a fast-lane package swap whose recorded constraints (#292) are the tool choice, the packaging source, the GPU backend, no always-on service, no cleanup, no models.

Worth keeping on the record: name resolution *is* falsifiable today — you verified a bogus name fails the build, and the check-mode diff pins the exact package set (`+llama-cpp`, `+ggml-hip`, `+ggml-cpu`) including the ROCm dependency closure. So the gate catches a package that disappears or gets renamed away; it does not catch a package that resolves and ships broken binaries. That residual is precisely what #301 names, with your `llama-cli --version` evidence and `llama-server --help` as the assertions to add in a real-install stage.

Resolving here; #301 is the open record.

PR #300, opened 2026-08-17
feat(home): auto-bind Copilot key (F23) to the Claude Code launcher on KDE hosts

### LINKED ISSUE #297: Copilot key launches Claude Code automatically after provisioning
## Problem statement

Pressing the Copilot key on the provisioned laptop does nothing. The hardware layer works — keyd remaps the physical Copilot chord to a bare `F23` keypress (PR #286) — and the home role deploys a Claude Code launcher (`hanzo-claude-code.desktop`), but nothing connects the two: PR #289 removed the KDE global-shortcut binding, leaving it as a manual System Settings step on each host. That defeats the point of a provisioner: after a fresh provision the key is dead until the operator configures it by hand.

Separately, the launcher currently runs `claude agents`; the desired command is plain `claude`.

## Expected behavior

After hanzo provisions a host with KDE Plasma installed — with or without an active Plasma session at provision time — and at most one logout/login, pressing a key that emits `F23` opens a Ghostty window in `~/cowork` running `claude`. The binding is established by the playbook itself — no manual System Settings step — and keeps working across KDE session restarts and repeated provisioning runs.

## Constraints

- The desktop entry keeps launching through the Ghostty wrapper; the launched command is plain `claude` (not `claude agents`); the `Comment` line stays exactly `Open Claude Code in ~/cowork` — user directives. The requirement is the outcome (Ghostty, `~/cowork`, plain `claude`); the exact Exec syntax is the implementation's to choose and must be valid per the Desktop Entry Specification (note: the current `%h` is not a valid field code).
- The binding must be fully automatic after provisioning; a documented manual step is not an acceptable outcome.
- A one-time logout/login (or Plasma restart) after a run that establishes or re-establishes the binding is acceptable for activation; between such runs the binding must persist without further action.
- "KDE host" means Plasma is installed on the machine. The binding is established even when provisioning runs without an active session (SSH/TTY): config lands in the user's home and activates at the next login.
- The KDE-side binding applies on any KDE host hanzo provisions, not only the GZ302; only the physical Copilot-chord remap remains GZ302-gated in the hardware role.
- PR #289's failure claims (session-start rewrite of `kglobalshortcutsrc`, no reload path, Settings invisibility) are unverified hypotheses, not ground truth: PR #280 reported the simple `kglobalshortcutsrc` write working on-device after one login, and KDE documentation contradicts parts of #289. Do not rule out the simple mechanism on #289's word alone.

## Goals

1. **G-1** — Pressing the Copilot key (F23) launches Claude Code on a provisioned KDE host with no manual configuration step.
2. **G-2** — The launcher opens a Ghostty window in `~/cowork` running plain `claude`.
3. **G-3** — The binding is durable: it survives KDE session restarts and repeated playbook runs, and re-runs are idempotent.

## Non-goals

- No change to the hardware-side keyd remap (Copilot chord → F23); it stays as merged in PR #286.
- The shortcut does not need to be visible or editable in the System Settings → Shortcuts GUI; only the key working is required.
- No support for desktop environments other than KDE Plasma; on non-KDE hosts the binding is simply not established.
- No alternative launch surfaces (menu pinning, taskbar, autostart) — only the key binding.

## Acceptance criteria

1. **AC-1** (G-1) — On a KDE host, after a full provisioning run and at most one logout/login, pressing a key that emits `F23` opens the launcher; no System Settings interaction has occurred. Proven by on-device demonstration.
2. **AC-2** (G-2) — The deployed `~/.local/share/applications/hanzo-claude-code.desktop` launches through Ghostty, targets `~/cowork` as working directory, and runs plain `claude` (no `agents` argument), with a spec-valid Exec line. Proven by reading the deployed file.
3. **AC-3** (G-3) — A second provisioning run on an already-configured host reports the binding tasks as unchanged. Proven by the playbook's changed-count for those tasks.
4. **AC-4** (G-3) — After a further logout/login beyond the activation one, the key still launches Claude Code. Proven by on-device demonstration.
5. **AC-5** (G-2) — The window opened by the keypress has working directory `~/cowork` and a running `claude` process. Proven by on-device demonstration.

## Edge cases

1. **EC-1** — Host without KDE installed / the CachyOS test container: the binding tasks skip cleanly and the full container check (`docker build -f tests/Containerfile -t hanzo:test .`) passes.
2. **EC-2** — Fresh KDE home where the relevant KDE config files do not exist yet, provisioned over SSH with no active session: the run succeeds and the binding activates after the first login.
3. **EC-3** — The operator has manually bound F23 to something else in KDE: the next provisioning run re-establishes the Claude Code binding, effective at the next logout/login (same activation allowance as first provision). Until that run, double-activation of both bindings on one keypress is an accepted interim state.

## Open questions & assumptions

- Assumption: each keypress opens a new Ghostty+claude session, even if one is already open — current launcher semantics; focus-existing is out of scope. If wrong, that is a follow-up change.
- Assumption: hanzo owns the F23 binding on provisioned hosts; conflicting operator-assigned bindings are overwritten on the next run (EC-3).
- Dependency: the Copilot chord → F23 keyd remap (roles/hardware, PR #286) is merged and working on the GZ302.
- Dependency: target hosts run KDE Plasma 6.
- Dependency (for implementation to verify): `claude` is installed via fnm-managed npm, so its PATH visibility in a desktop-activation context (no shell rc) must be confirmed on-device — AC-5 catches a failure here.

palazzem:
**Claude Harness**

Restores the automatic KDE-side binding of `F23` to the Claude Code launcher (removed in #289) so the Copilot key works after a provisioning run with no System Settings step, and fixes the launcher's `Exec` line: it is now spec-valid, opens Ghostty in `~/cowork`, and runs plain `claude`. The binding is written to the group kglobalacceld actually reads on Plasma 6 — the nested `[services][hanzo-claude-code.desktop]` group with a bare key sequence — which is why the pre-#289 top-level `[hanzo-claude-code.desktop]` group never took effect.

## Run record

- **Issue:** #297
- **Session:** `ea4b2d65-17c6-4981-aaf1-cd3ce4a7a8af` — resume with `claude --resume ea4b2d65-17c6-4981-aaf1-cd3ce4a7a8af`

### Related Issues

- fixes #297

### Proposed Changes:

**Detection fact (`playbook.yml` pre_tasks)** — new `host_has_kde`, stat of `/usr/bin/plasmashell`. "KDE host" means Plasma is *installed*, not that a session is running, so provisioning over SSH/TTY still writes the config; session-scoped signals (`KDE_FULL_SESSION`, `XDG_CURRENT_DESKTOP`) would wrongly skip the binding on a headless run. Per CLAUDE.md rule 3 the fact lives in `pre_tasks`; the role only consumes it.

**KDE binding (`roles/home/tasks/kde.yml`)** — one `community.general.ini_file` task, gated on `host_has_kde`, writing `_launch=F23` into `~/.config/kglobalshortcutsrc`. The KDE user-side tasks (launcher entry + binding) live in their own `kde.yml`, included unconditionally from `roles/home/tasks/main.yml`, so further KDE settings have an obvious home.

Why the group changed shape vs. the pre-#289 task — read from kglobalacceld 6.7.4 (the version Arch/CachyOS ships, `extra/kglobalacceld 6.7.4-1`):

- `GlobalShortcutsRegistry::loadSettings()` iterates top-level groups and **skips** any group ending in `.desktop`, then loads launcher shortcuts from `_config.group("services").group(<name>.desktop)` — i.e. the INI group `[services][hanzo-claude-code.desktop]`.
- `GlobalShortcutsRegistry::migrateConfig()` runs at daemon construction and **deletes** top-level `<name>.desktop` groups, migrating only values that are a valid `shortcut,default,friendly` triplet.
- The old task wrote a top-level `[hanzo-claude-code.desktop]` group with `_launch=F23` (not a triplet), so it was dropped on the first daemon start and never registered. That is the concrete reason the pre-#289 binding did not stick — not #289's stated causes.
- `KServiceActionComponent::loadSettings()` reads `_launch` as a bare key sequence (`configGroup.readEntry("_launch", …)`), so the value is `F23`, not a triplet.
- Durability: `KServiceActionComponent::writeSettings()` only drops the entry when the configured shortcut equals the desktop file's default (`X-KDE-Shortcuts`). We deliberately do *not* set `X-KDE-Shortcuts`, so `F23` differs from the (empty) default and KDE re-persists it on every rewrite instead of reverting it.

`ini_file` has no nested-group syntax, hence the `section: "services][{{ home_claude_launcher_file }}"` spelling; a short comment in the task records that, with the full derivation kept here rather than in the source.

**Launcher basename var (`roles/home/vars/main.yml`, commit c248670)** — the basename is load-bearing twice (kglobalacceld resolves the shortcut to the launcher *by filename*, so the `dest:` of the template task and the `section:` of the binding task must stay byte-identical). Both now read `home_claude_launcher_file` instead of independent literals, following the role's existing `home_gitidentity_file` convention (var + warning comment for cross-file couplings that fail silently). Answers `settling/F1`.

**Launcher entry (`roles/home/templates/hanzo-claude-code.desktop.j2`)** — `Exec=ghostty --working-directory={{ ansible_facts.env.HOME }}/cowork -e fish -c claude`. `%h` is not a Desktop Entry field code (only `%f %F %u %U %i %c %k`), so the home path is templated in. `agents` is dropped. `claude` is an npm global under the fnm-managed Node version and is only on PATH once the shell environment loads (dotfiles `conf.d/environment.fish`, sourced by fish for every session kind); a desktop activation inherits the session environment, which has no fnm PATH entry — hence `fish -c claude` rather than a bare `claude`. `Comment` is unchanged.

### Testing:

**Lint** — `pre-commit run --all-files`: all hooks pass (ansible-lint included).

**Full container check** — `docker build -f tests/Containerfile -t hanzo:test .`: passes, `failed=0`. EC-1 confirmed — on the non-KDE container the binding task reports `skipping: [localhost]`, and the rendered entry is:

```
TASK [home : Deploy Claude Code launcher .desktop entry] ***********************
+Exec=ghostty --working-directory=/home/testuser/cowork -e fish -c claude
TASK [home : Bind Copilot key (F23) to Claude Code launcher in KDE] ************
skipping: [localhost]
```

**AC-2 — spec-valid Exec** (`desktop-file-validate`, in-container):

```
$ desktop-file-validate hanzo-claude-code.desktop        # new entry
(no errors; one "more than one main category" hint, pre-existing)
$ desktop-file-validate old.desktop                      # entry as on main
error: value "ghostty --working-directory=%h/cowork -e claude agents" for key "Exec"
       in group "Desktop Entry" contains an invalid field code "%h"
exit=1
```

**AC-3 — idempotency of the binding task** (module-level, in-container):

```
run 1: localhost | CHANGED => { "changed": true,
run 2: localhost | SUCCESS => { "changed": false,
file:  [services][hanzo-claude-code.desktop]
       _launch=F23
```

**EC-3 — operator rebound F23 elsewhere**: seeding `_launch=Meta+X` and re-running the task reports `"changed": true` and restores `_launch=F23`.

**EC-2 — fresh KDE home, no existing KDE config, no active session** (in-container, `HOME` with no `.config` directory at all; task parameters taken from the role, role vars loaded so both use sites render from `home_claude_launcher_file`):

```
--- fresh HOME, no ~/.config:
=== EC-2 run 1 (fresh home) ===
    "msg": [
        "dest:    /tmp/ec2home/.local/share/applications/hanzo-claude-code.desktop",
        "section: services][hanzo-claude-code.desktop"
    ]
TASK [Bind Copilot key (F23) to Claude Code launcher in KDE] *******************
changed: [localhost]
localhost : ok=3  changed=1  unreachable=0  failed=0
=== EC-2 run 2 (idempotency) ===
localhost : ok=3  changed=0  unreachable=0  failed=0
=== resulting file ===
drwxr-xr-x 1 testuser testuser 36 /tmp/ec2home/.config
-rw-r--r-- 1 testuser testuser 51 /tmp/ec2home/.config/kglobalshortcutsrc

[services][hanzo-claude-code.desktop]
_launch=F23
```

`ini_file` creates the missing `~/.config` parent itself, so the first run on a fresh home succeeds and the binding is on disk for the next login. The same run is the evidence that the two use sites of `home_claude_launcher_file` render identically.

**AC-1 / AC-4 / AC-5 — on-device, for the operator** (not claimed here; the spec's proof channel for these is on-device demonstration, which cannot be produced from CI or the container — these boxes are the merge gate):

- [ ] Run `hanzo` on the GZ302, then log out and back in to KDE once.
- [ ] **AC-1** — press the Copilot key: a Ghostty window opens.
- [ ] **AC-5** — in that window, `pwd` is `~/cowork` and `claude` is running.
- [ ] **AC-4** — log out and back in again; the key still launches.

If the AC-5 step fails with `claude: command not found`, the fnm PATH assumption in the `Exec` line needs revisiting — report the exact error.

Optional check without a logout: `grep -A1 'services\]\[hanzo-claude-code.desktop' ~/.config/kglobalshortcutsrc` should show `_launch=F23`.

### Extra Notes (optional):

Review focus: the `section: "services][{{ home_claude_launcher_file }}"` spelling in `roles/home/tasks/kde.yml` (it looks like a typo but is the only way `ini_file` can express a KConfig nested group), and the `fish -c` wrapper in the Exec line.

### Checklist

### DIFF
--- a/roles/home/tasks/main.yml
+++ b/roles/home/tasks/main.yml
@@ -76,3 +76,24 @@
     dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/hanzo-claude-code.desktop"
     mode: "0644"
   become: false
+
+# kglobalacceld (Plasma 6) reads launcher shortcuts from the *nested*
+# `[services][<file>.desktop]` group of kglobalshortcutsrc, with the bare
+# key sequence as value — see kglobalacceld's GlobalShortcutsRegistry::
+# loadSettings() and KServiceActionComponent::loadSettings(). Top-level
+# `[<file>.desktop]` groups — the shape used before #289 — are skipped by
+# loadSettings() and deleted by migrateConfig() on every daemon start,
+# which is why that earlier write never took effect. ini_file has no
+# nested-group syntax, hence the `]`+`[` embedded in `section`.
+# Writing this from an SSH/TTY run is fine: kglobalacceld reads the file
+# at session start, so the binding activates at the next login.
+- name: Bind Copilot key (F23) to Claude Code launcher in KDE
+  community.general.ini_file:
+    path: "{{ ansible_facts.env.HOME }}/.config/kglobalshortcutsrc"
+    section: "services][hanzo-claude-code.desktop"
+    option: _launch
+    value: "F23"
+    no_extra_spaces: true
+    mode: "0644"
+  when: host_has_kde
+  become: false
--- a/roles/home/tasks/main.yml
+++ b/roles/home/tasks/main.yml
@@ -60,42 +60,5 @@
     mode: "0755"
   become: false
 
-# ansible.builtin.template does not create parent directories.
-# On a fresh KDE setup ~/.local/share/applications may not exist
-# yet — pre-create it so the template task is robust on first run.
-- name: Ensure ~/.local/share/applications exists
-  ansible.builtin.file:
-    path: "{{ ansible_facts.env.HOME }}/.local/share/applications"
-    state: directory
-    mode: "0755"
-  become: false
-
-- name: Deploy Claude Code launcher .desktop entry
-  ansible.builtin.template:
-    src: hanzo-claude-code.desktop.j2
-    dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/{{ home_claude_launcher_file }}"
-    mode: "0644"
-  become: false
-
-# kglobalacceld (Plasma 6) reads launcher shortcuts from the *nested*
-# `[services][<file>.desktop]` group of kglobalshortcutsrc, with the bare
-# key sequence as value — see kglobalacceld's GlobalShortcutsRegistry::
-# loadSettings() and KServiceActionComponent::loadSettings(). Top-level
-# `[<file>.desktop]` groups — the shape used before #289 — are skipped by
-# loadSettings() and deleted by migrateConfig() on every daemon start,
-# which is why that earlier write never took effect. ini_file has no
-# nested-group syntax, hence the `]`+`[` embedded in `section`.
-# The launcher basename comes from home_claude_launcher_file — the same
-# var the template task deploys under, since the two must match exactly.
-# Writing this from an SSH/TTY run is fine: kglobalacceld reads the file
-# at session start, so the binding activates at the next login.
-- name: Bind Copilot key (F23) to Claude Code launcher in KDE
-  community.general.ini_file:
-    path: "{{ ansible_facts.env.HOME }}/.config/kglobalshortcutsrc"
-    section: "services][{{ home_claude_launcher_file }}"
-    option: _launch
-    value: "F23"
-    no_extra_spaces: true
-    mode: "0644"
-  when: host_has_kde
-  become: false
+- name: Configure KDE Plasma user settings
+  ansible.builtin.include_tasks: kde.yml
--- a/roles/home/tasks/main.yml
+++ b/roles/home/tasks/main.yml
@@ -73,7 +73,7 @@
 - name: Deploy Claude Code launcher .desktop entry
   ansible.builtin.template:
     src: hanzo-claude-code.desktop.j2
-    dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/hanzo-claude-code.desktop"
+    dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/{{ home_claude_launcher_file }}"
     mode: "0644"
   become: false
 
@@ -85,12 +85,14 @@
 # loadSettings() and deleted by migrateConfig() on every daemon start,
 # which is why that earlier write never took effect. ini_file has no
 # nested-group syntax, hence the `]`+`[` embedded in `section`.
+# The launcher basename comes from home_claude_launcher_file — the same
+# var the template task deploys under, since the two must match exactly.
 # Writing this from an SSH/TTY run is fine: kglobalacceld reads the file
 # at session start, so the binding activates at the next login.
 - name: Bind Copilot key (F23) to Claude Code launcher in KDE
   community.general.ini_file:
     path: "{{ ansible_facts.env.HOME }}/.config/kglobalshortcutsrc"
-    section: "services][hanzo-claude-code.desktop"
+    section: "services][{{ home_claude_launcher_file }}"
     option: _launch
     value: "F23"
     no_extra_spaces: true
--- a/roles/home/templates/hanzo-claude-code.desktop.j2
+++ b/roles/home/templates/hanzo-claude-code.desktop.j2
@@ -1,14 +1,3 @@
-{#
-  Exec notes:
-  - The home path is templated in because %h is not a Desktop Entry field
-    code (only %f %F %u %U %i %c %k are) and would be passed through
-    literally by the launcher.
-  - `claude` is an npm global under the fnm-managed Node version, so it is
-    only on PATH once the shell environment is loaded (dotfiles
-    conf.d/environment.fish, which fish sources for every session kind).
-    A desktop activation starts from the session environment, which has no
-    fnm PATH entry — hence `fish -c claude` instead of a bare `claude`.
--#}
 [Desktop Entry]
 Type=Application
 Name=Claude Code
--- a/roles/home/templates/hanzo-claude-code.desktop.j2
+++ b/roles/home/templates/hanzo-claude-code.desktop.j2
@@ -1,8 +1,19 @@
+{#
+  Exec notes:
+  - The home path is templated in because %h is not a Desktop Entry field
+    code (only %f %F %u %U %i %c %k are) and would be passed through
+    literally by the launcher.
+  - `claude` is an npm global under the fnm-managed Node version, so it is
+    only on PATH once the shell environment is loaded (dotfiles
+    conf.d/environment.fish, which fish sources for every session kind).
+    A desktop activation starts from the session environment, which has no
+    fnm PATH entry — hence `fish -c claude` instead of a bare `claude`.
+-#}
 [Desktop Entry]
 Type=Application
 Name=Claude Code
 Comment=Open Claude Code in ~/cowork
-Exec=ghostty --working-directory=%h/cowork -e claude agents
+Exec=ghostty --working-directory={{ ansible_facts.env.HOME }}/cowork -e fish -c claude
 Icon=utilities-terminal
 Terminal=false
 Categories=Utility;Development;
--- a/playbook.yml
+++ b/playbook.yml
@@ -31,12 +31,26 @@
       become: false
       tags: [always]
 
+    # "KDE host" means Plasma is *installed*, not that a session is
+    # running: provisioning over SSH/TTY must still write the KDE config
+    # that activates at the next login. /usr/bin/plasmashell is the
+    # install marker; session-scoped signals (KDE_FULL_SESSION,
+    # XDG_CURRENT_DESKTOP) are absent on a headless run and would wrongly
+    # skip the binding.
+    - name: Detect installed KDE Plasma
+      ansible.builtin.stat:
+        path: /usr/bin/plasmashell
+      register: hanzo_plasma_stat
+      become: false
+      tags: [always]
+
     # `'product_name' in ansible_facts` short-circuits the GZ302 check
     # so it's safe even when the DMI fact is missing (gather_subset
     # without `hardware`, or non-Linux DMI paths).
     - name: Compute provisioning facts
       ansible.builtin.set_fact:
         host_has_systemd: "{{ hanzo_systemd_stat.stat.exists }}"
+        host_has_kde: "{{ hanzo_plasma_stat.stat.exists }}"
         model_is_gz302: "{{ 'product_name' in ansible_facts
                            and 'GZ302' in ansible_facts['product_name']
                            and hanzo_systemd_stat.stat.exists }}"
--- a/roles/home/tasks/kde.yml
+++ b/roles/home/tasks/kde.yml
@@ -1,0 +1,35 @@
+---
+# KDE Plasma user-side configuration: files that live in the user's home
+# and are read by Plasma at session start. Writing them from an SSH/TTY
+# run is fine — they activate at the next login.
+
+# ansible.builtin.template does not create parent directories, and a
+# fresh home has no ~/.local/share/applications yet.
+- name: Ensure ~/.local/share/applications exists
+  ansible.builtin.file:
+    path: "{{ ansible_facts.env.HOME }}/.local/share/applications"
+    state: directory
+    mode: "0755"
+  become: false
+
+- name: Deploy Claude Code launcher .desktop entry
+  ansible.builtin.template:
+    src: hanzo-claude-code.desktop.j2
+    dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/{{ home_claude_launcher_file }}"
+    mode: "0644"
+  become: false
+
+# kglobalacceld reads launcher shortcuts only from the nested
+# `[services][<file>.desktop]` group; top-level `[<file>.desktop]` groups
+# are ignored and deleted on daemon start. ini_file has no nested-group
+# syntax, hence the `]`+`[` embedded in `section`.
+- name: Bind Copilot key (F23) to Claude Code launcher in KDE
+  community.general.ini_file:
+    path: "{{ ansible_facts.env.HOME }}/.config/kglobalshortcutsrc"
+    section: "services][{{ home_claude_launcher_file }}"
+    option: _launch
+    value: "F23"
+    no_extra_spaces: true
+    mode: "0644"
+  when: host_has_kde
+  become: false
--- a/roles/home/vars/main.yml
+++ b/roles/home/vars/main.yml
@@ -11,3 +11,11 @@ home_user_services:
 # Path must match `[include] path = ...` in palazzem/dotfiles gitconfig.
 # Renaming this breaks the cross-repo include and silently drops identity.
 home_gitidentity_file: ~/.gitidentity
+
+# Basename of the Claude Code launcher entry. kglobalacceld resolves the
+# global shortcut to the launcher *by this filename*: the deployed file in
+# ~/.local/share/applications and the `[services][<file>]` group in
+# kglobalshortcutsrc must stay byte-identical, so both use this var.
+# Renaming it in only one place leaves the shortcut pointing at a
+# nonexistent service and the Copilot key silently goes dead.
+home_claude_launcher_file: hanzo-claude-code.desktop
--- a/roles/home/vars/main.yml
+++ b/roles/home/vars/main.yml
@@ -12,10 +12,7 @@ home_user_services:
 # Renaming this breaks the cross-repo include and silently drops identity.
 home_gitidentity_file: ~/.gitidentity
 
-# Basename of the Claude Code launcher entry. kglobalacceld resolves the
-# global shortcut to the launcher *by this filename*: the deployed file in
-# ~/.local/share/applications and the `[services][<file>]` group in
-# kglobalshortcutsrc must stay byte-identical, so both use this var.
-# Renaming it in only one place leaves the shortcut pointing at a
-# nonexistent service and the Copilot key silently goes dead.
+# KDE resolves the global shortcut to the launcher by this filename, so the
+# deployed .desktop file and the kglobalshortcutsrc group must match exactly.
+# Renaming it in only one place silently kills the Copilot key.
 home_claude_launcher_file: hanzo-claude-code.desktop


palazzem:
**Claude Harness**

Review-panel angle selection for this PR (deterministic rubric):

**Core — always on:**
- `correctness` — selected
- `divergence` — selected (spec issue #297; no design doc — fast lane)
- `simplicity` — selected
- `settling` — selected

**Conditional:**
- `security` — skipped: diff touches no authn/authz, secret handling, crypto, or input parsing/deserialization
- `reliability` — skipped: no new/changed call across a process boundary or retry/timeout/recovery logic in provisioning code
- `performance` — skipped: no loop-body changes, no caching

**Repo-supplied:** none (`.claude/review-angles/` absent)

palazzem (commented):
**Claude Harness**

Examined every hunk of the PR against the charter "abstraction, indirection, or generality no requirement demands": the `hanzo_plasma_stat` stat task and `host_has_kde` fact in `playbook.yml` pre_tasks; the `community.general.ini_file` binding task in `roles/home/tasks/main.yml` (including the `services][hanzo-claude-code.desktop` section spelling and the `no_extra_spaces`/`mode` options); and the reworked `Exec` line in `roles/home/templates/hanzo-claude-code.desktop.j2` (`{{ ansible_facts.env.HOME }}` templating and the `fish -c claude` wrapper). Each layer traces to a stated demand: the pre_tasks fact shape is mandated by CLAUDE.md rule 3 and mirrors the existing `host_has_systemd` pattern; the single `ini_file` call is the minimal idempotent write for a KConfig nested group (no shell escape hatch, no whole-file template); the `fish -c` wrapper answers the spec's named fnm-PATH dependency, where the less-indirect absolute path would pin a Node-version directory; the HOME templating is forced by `%h` being an invalid Desktop Entry field code (AC-2) in a file that was already a deployed `.j2`. The KDE-wide (non-GZ302) scope of the binding is explicitly required by the spec. No unused generality, speculative options, or added machinery found; zero findings.

palazzem (commented):
**Claude Harness**

**Angle `settling` — 1 finding(s)**

Examined the three changed files for places where the diff carries a pre-existing pattern forward despite an evidently better approach: the stat-detection fact pattern in `playbook.yml` pre_tasks (clean copy of the mandated `hanzo_systemd_stat` convention), the `ansible_facts.env.HOME` path convention, the `fish -c` PATH strategy in the Exec line, the `]`-embedding in the `ini_file` `section` (verified against the installed `community.general` module source: section matching is a `re.escape`d literal with no nested-group facility, so this is the only expressible form — not a settled-for hack), and the restored pre-#289 shape of the binding task. One finding.

- `settling/F1` — the load-bearing launcher basename `hanzo-claude-code.desktop` is duplicated as bare literals across two coupled tasks, restoring the pre-#289 shape, where the role's own var-with-warning convention is the evident better approach → roles/home/tasks/main.yml:93

roles/home/tasks/main.yml
```diff
@@ -76,3 +76,24 @@
...
+# `[<file>.desktop]` groups — the shape used before #289 — are skipped by
+# loadSettings() and deleted by migrateConfig() on every daemon start,
+# which is why that earlier write never took effect. ini_file has no
+# nested-group syntax, hence the `]`+`[` embedded in `section`.
+# Writing this from an SSH/TTY run is fine: kglobalacceld reads the file
+# at session start, so the binding activates at the next login.
+- name: Bind Copilot key (F23) to Claude Code launcher in KDE
+  community.general.ini_file:
+    path: "{{ ansible_facts.env.HOME }}/.config/kglobalshortcutsrc"
+    section: "services][hanzo-claude-code.desktop"
```
palazzem: **Claude Harness**

**`settling/F1` — The diff restores the pre-#289 convention of duplicating the load-bearing basename `hanzo-claude-code.desktop` as independent bare literals in two coupled tasks, when the role's own var convention is the evident better approach**

- **Evidence:** Traced path. (1) `roles/home/tasks/main.yml:76` deploys the launcher to `dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/hanzo-claude-code.desktop"`; line 93 independently spells the same basename in `section: "services][hanzo-claude-code.desktop"`. Per this task's own comment, kglobalacceld resolves the binding to the launcher *by that filename* (`KServiceActionComponent::loadSettings()` reads `[services][<file>.desktop]`), so the two literals must stay byte-identical for the key to work at all. (2) This is exactly the shape #289 deleted: `git show 98aa1ea -- roles/home/tasks/main.yml` shows the removed task with the same duplicated literal (`section: hanzo-claude-code.desktop` alongside the same `dest:`) — the duplication is carried forward on reintroduction, not fixed. (3) The repo itself names the better approach for this class of coupling: `roles/home/vars/main.yml` defines `home_gitidentity_file` with the comment "Path must match `[include] path = ...` in palazzem/dotfiles gitconfig. Renaming this breaks the cross-repo include and silently drops identity" — one role var, consumed at every use site. (4) No automated check covers drift between the two literals: per EC-1 the binding task reports `skipping: [localhost]` in the only test environment (the non-KDE container), so a rename touching one site but not the other passes `pre-commit run --all-files` and the full container build.
- **Impact:** A future rename of the launcher file (or edit of the section string) that updates one site but not the other leaves `kglobalshortcutsrc` pointing at a nonexistent service: the Copilot key silently goes dead — the exact regression class this PR exists to close (#297) — and nothing short of on-device testing can detect it. A single `home_claude_launcher_file`-style var referenced from both `dest:` and `section:` removes the failure mode outright.

roles/home/tasks/main.yml
```diff
@@ -76,3 +76,24 @@
...
+# `[services][<file>.desktop]` group of kglobalshortcutsrc, with the bare
+# key sequence as value — see kglobalacceld's GlobalShortcutsRegistry::
+# loadSettings() and KServiceActionComponent::loadSettings(). Top-level
+# `[<file>.desktop]` groups — the shape used before #289 — are skipped by
+# loadSettings() and deleted by migrateConfig() on every daemon start,
+# which is why that earlier write never took effect. ini_file has no
+# nested-group syntax, hence the `]`+`[` embedded in `section`.
+# Writing this from an SSH/TTY run is fine: kglobalacceld reads the file
+# at session start, so the binding activates at the next login.
+- name: Bind Copilot key (F23) to Claude Code launcher in KDE
```
palazzem: **Claude Harness**

**`divergence/F1` — Three of the spec's five acceptance criteria (AC-1, AC-4, AC-5) ship unproven: their spec-mandated on-device demonstration is deferred to a post-merge operator checklist with no tracking mechanism**

- **Evidence:** Spec #297 fixes the proof mode for each criterion — AC-1: "Proven by on-device demonstration.", AC-4: "Proven by on-device demonstration.", AC-5: "Proven by on-device demonstration." The PR body declares the opposite: "**AC-1 / AC-4 / AC-5 — on-device, for the operator** (not claimed here)". This binding task is the delivery point for AC-1/AC-4; the deployed `Exec` line for AC-5. The mechanism argument itself checks out — I independently verified the kglobalacceld v6.7.4 claims against upstream source (`loadSettings()` skips top-level groups via `if (groupName.endsWith(QLatin1String(".desktop"))) { continue; }` and loads launchers from `_config.group(QStringLiteral("services"))`; `migrateConfig()` deletes top-level `.desktop` groups and migrates only 3-part triplets; `KServiceActionComponent` reads `_launch` as a bare sequence and drops it only when equal to the `X-KDE-Shortcuts` default, which this entry deliberately omits) and the `fish -c claude` PATH claim against `palazzem/dotfiles` `conf.d/environment.fish` (unconditional `fnm env --use-on-cd --shell fish | source`, no interactivity gate) — but a source-verified argument is not the proof the spec demands, and the spec itself flags the residual risk as on-device-only: "Dependency (for implementation to verify): `claude` is installed via fnm-managed npm, so its PATH visibility in a desktop-activation context (no shell rc) must be confirmed on-device — AC-5 catches a failure here."
- **Impact:** The PR can merge with the end-to-end outcome of G-1/G-2 (keypress → Ghostty in `~/cowork` running `claude`) never demonstrated on the target host; if the operator's step 3 fails (`claude: command not found` or the binding not registering), the failure surfaces after merge with nothing tracking the unmet criteria.

palazzem (commented):
**Claude Harness**

**Angle `divergence` — 1 finding(s)**

Compared the three changed files against spec issue #297 (no approved design exists; the PR's own declarations served as the secondary reference): the `host_has_kde` fact in `playbook.yml` pre_tasks vs the spec's "KDE host means Plasma is installed" constraint; the `ini_file` binding task in `roles/home/tasks/main.yml` vs G-1/G-3, AC-3, and EC-1–EC-3; the `Exec` rewrite in `roles/home/templates/hanzo-claude-code.desktop.j2` vs the Ghostty/`~/cowork`/plain-`claude`/spec-valid-Exec constraint and AC-2. Verified the PR's kglobalacceld mechanism claims against upstream v6.7.4 source, its `fish -c` PATH claim against `palazzem/dotfiles` `conf.d/environment.fish`, and EC-2 fresh-home handling against community.general's `ini_file` module source (creates missing parent directories). All of those hold; the one divergence found is the spec's proof requirement.

- `divergence/F1 — AC-1/AC-4/AC-5's spec-mandated on-device proof is not delivered; three of five acceptance criteria ship as a post-merge operator checklist → roles/home/tasks/main.yml:90`

palazzem (commented):
**Claude Harness**

Examined the three changed files (`playbook.yml` pre_tasks, `roles/home/tasks/main.yml`, `roles/home/templates/hanzo-claude-code.desktop.j2`) against the PR's behavioral claims, re-verifying each in an isolated worktree of the branch. (1) Container check-mode build reproduced: `failed=0`, binding task `skipping` on the non-KDE container (EC-1), rendered Exec matches the claim. (2) `community.general.ini_file` traced at source: the section regex `^\[\s*{re.escape(section)}\s*]` matches the literal `[services][hanzo-claude-code.desktop]` line, so the `][` spelling is a real nested-group write and re-match, and `os.makedirs(destpath)` covers a missing `~/.config` (EC-2). (3) In-container module runs against a pre-populated `kglobalshortcutsrc` (top-level groups plus a foreign `[services][*.desktop]` subgroup): run 1 `changed`, run 2 unchanged (AC-3), seeded `_launch=Meta+X` restored to `F23` (EC-3), neighboring groups untouched. (4) kglobalacceld v6.7.4 source confirms the mechanism: `loadSettings()` skips top-level `.desktop` groups and reads `_config.group("services").group(<name>.desktop)`; `migrateConfig()` deletes top-level `.desktop` groups and migrates only valid triplets; `KServiceActionComponent` reads `_launch` as a bare key sequence and `writeSettings()` re-persists non-default shortcuts — so `F23` with no `X-KDE-Shortcuts` default survives daemon rewrites. (5) Exec chain: ghostty docs confirm `-e` treats all trailing args as the command (`fish -c claude`); dotfiles `conf.d/environment.fish` runs `fnm env … | source` unconditionally for every fish session kind, and `fnm` is `/usr/bin/fnm` via `fnm-bin`, so `claude` resolves in a desktop-activation context. (6) Rendered entry passes `desktop-file-validate` (exit 0, pre-existing category hint only). No logic, boundary, or error-path defect found; the on-device criteria (AC-1/4/5) are correctly delegated to the operator, not claimed.

roles/home/tasks/main.yml
```diff
@@ -76,3 +76,24 @@
...
+  community.general.ini_file:
+    path: "{{ ansible_facts.env.HOME }}/.config/kglobalshortcutsrc"
```
palazzem: **Claude Harness**

**`testing-quality/F1` — Spec EC-2 (fresh KDE home, no existing config files) has no test evidence anywhere in the PR**

- **Evidence:** Spec #297 EC-2: "Fresh KDE home where the relevant KDE config files do not exist yet, provisioned over SSH with no active session: the run succeeds and the binding activates after the first login." The PR's Testing section enumerates EC-1, AC-2, AC-3, EC-3 and explicitly defers AC-1/AC-4/AC-5 to the operator — EC-2 appears nowhere: neither tested nor declared deferred. The AC-3 module-level run does not stand in for it, because the test container pre-seeds `/home/testuser/.config/hanzo/config.yml` (`tests/Containerfile` lines 29–31), so `~/.config` already existed there; EC-2's distinguishing condition — no `~/.config` directory at all — was never exercised. Reproduction in an isolated container (exact task parameters, `HOME` containing no `.config`): run 1 `ok=1 changed=1 failed=0`, run 2 `changed=0`, file created with exactly `[services][hanzo-claude-code.desktop]` / `_launch=F23` — so the case passes today; the gap is evidential, not behavioral.
- **Impact:** The fresh-home path relies on `community.general.ini_file` silently creating the missing parent directory (`os.makedirs(destpath)` in the module's create-branch — undocumented behavior). With no EC-2 evidence in the run record and the checklist marked "Tests are defining the correct and expected behavior", a regression in that path (or a module behavior change) ships green while a fresh host provisioned over SSH fails its first run — exactly the scenario EC-2 exists to gate.

palazzem (commented):
**Claude Harness**

This angle examined whether the PR's test evidence proves its claims against spec #297's ACs/ECs, across `playbook.yml` (the `host_has_kde` pre_task), `roles/home/tasks/main.yml` (the `ini_file` binding task), and `roles/home/templates/hanzo-claude-code.desktop.j2` (the Exec line). Every claim in the Testing section was re-run in an isolated container: the full container check (EC-1 — binding task `skipping: [localhost]`, `failed=0`, rendered `Exec=ghostty --working-directory=/home/testuser/cowork -e fish -c claude`); `desktop-file-validate` on the new entry (exit 0) with the old `%h` entry as a negative control (exit 1 — the assertion can fail); module-level idempotency (run 1 `changed=1`, run 2 `changed=0`) on both a fresh file and a realistically populated `kglobalshortcutsrc` with no group duplication; the EC-3 override case (seeded `_launch=Meta+X` → `changed=1`, `F23` restored); both directions of the `host_has_kde` gate (task skips without `/usr/bin/plasmashell`, fires with it); and `pre-commit run --all-files` (all hooks pass). All reproduce; no tautological assertions found. Deferring AC-1/AC-4/AC-5 to the operator matches the spec's own proof channel ("Proven by on-device demonstration"). One spec edge case carries no evidence at all — the single finding below.

- `testing-quality/F1 — Spec EC-2 (fresh KDE home, no existing config files) has no test evidence anywhere in the PR → roles/home/tasks/main.yml:92`

roles/home/tasks/main.yml
palazzem: **Claude Harness**

**fix** — commit `c248670` ("refactor(home): single-source the Claude Code launcher basename").

The finding is right on both the coupling and the precedent: the basename is load-bearing twice (kglobalacceld resolves the shortcut to the launcher *by filename*), no automated check covers drift because the binding task skips in the only test environment (EC-1), and the role already names the better pattern with `home_gitidentity_file`.

What changed:

- `roles/home/vars/main.yml` — new `home_claude_launcher_file: hanzo-claude-code.desktop`, with the same style of warning comment as `home_gitidentity_file` ("Renaming it in only one place leaves the shortcut pointing at a nonexistent service and the Copilot key silently goes dead").
- `roles/home/tasks/main.yml:76` — `dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/{{ home_claude_launcher_file }}"`.
- `roles/home/tasks/main.yml:95` — `section: "services][{{ home_claude_launcher_file }}"`, plus a comment line pointing at the shared var.

Verified after the change:

- `pre-commit run --all-files` — all hooks pass (ansible-lint included).
- `docker build -f tests/Containerfile -t hanzo:test .` — `ok=59 changed=30 failed=0`; launcher entry still renders `Exec=ghostty --working-directory=/home/testuser/cowork -e fish -c claude`, binding task still `skipping: [localhost]` (EC-1 intact).
- Both use sites render identically from the var (in-container, role vars loaded):

```
"dest:    /tmp/ec2home/.local/share/applications/hanzo-claude-code.desktop"
"section: services][hanzo-claude-code.desktop"
```

and the written group is unchanged: `[services][hanzo-claude-code.desktop]` / `_launch=F23`.

The PR body's Proposed Changes section records the var.

roles/home/tasks/main.yml
palazzem: **Claude Harness**

**fix** — EC-2 evidence produced and recorded in the PR body's Testing section (the gap was evidential, as the finding states; no behavior change was needed, so this is a body/record fix rather than a code commit).

Reproduction, run in the CachyOS test image with a `HOME` that contains no `.config` directory at all — the condition `tests/Containerfile` lines 29–31 hide — using the role's own task parameters with `roles/home/vars/main.yml` loaded, so `dest:` and `section:` render from `home_claude_launcher_file` exactly as the role does (run against the post-`c248670` task):

```
--- fresh HOME, no ~/.config:
=== EC-2 run 1 (fresh home) ===
    "msg": [
        "dest:    /tmp/ec2home/.local/share/applications/hanzo-claude-code.desktop",
        "section: services][hanzo-claude-code.desktop"
    ]
TASK [Bind Copilot key (F23) to Claude Code launcher in KDE] *******************
changed: [localhost]
localhost : ok=3  changed=1  unreachable=0  failed=0
=== EC-2 run 2 (idempotency) ===
localhost : ok=3  changed=0  unreachable=0  failed=0
=== resulting file ===
drwxr-xr-x 1 testuser testuser 36 /tmp/ec2home/.config
-rw-r--r-- 1 testuser testuser 51 /tmp/ec2home/.config/kglobalshortcutsrc

[services][hanzo-claude-code.desktop]
_launch=F23
```

So EC-2's "the run succeeds" half is now evidenced: `ini_file` creates the missing `~/.config` parent, the first run on a fresh home is `changed=1 failed=0`, the second is `changed=0`, and the file on disk carries the group kglobalacceld reads at the next login (EC-2's activation half is the same on-device channel as AC-1 — see the `divergence/F1` thread).

On the residual point about `os.makedirs(destpath)` being undocumented module behavior: I did not add a permanent regression gate for it, because the only test environment this repo defines (the non-KDE container) skips the binding task by design (EC-1), so any such gate would have to be a task-level harness outside `tests/Containerfile` — new test infrastructure this issue does not ask for. If you want that gate, say so and I will file it as a follow-up rather than smuggle it into this PR.

roles/home/tasks/main.yml
palazzem: **Claude Harness**

**contest** — leaving this thread open for your ruling, because the decision it asks for (merge before the on-device demonstration, or hold) is yours, not mine.

Counter-evidence on the finding's two claims:

1. *"Ship unproven"* — the PR never claims them. AC-1/AC-4/AC-5 are marked "on-device, for the operator (not claimed here)"; the finding quotes that line as the divergence. The spec fixes the proof mode for those three as "Proven by on-device demonstration", and it also names the residual risk as on-device-only ("`claude` … PATH visibility in a desktop-activation context … must be confirmed on-device"). There is no channel in this run that can produce that proof: the only test environment the repo defines is the CachyOS container, which has no Plasma (`/usr/bin/plasmashell` absent → EC-1 skip) and no KDE session or physical F23 key; CI runs the same container; and CLAUDE.md forbids running `ansible`/`ansible-playbook`/`hanzo` on the host under any flag. The demonstration requires the GZ302 with a KDE login — the operator's hands. Producing anything labelled as AC-1/4/5 proof from here would be a fabricated claim, which is strictly worse than the honest deferral. The `testing-quality` angle read the same deferral as matching the spec's proof channel.

2. *"No tracking mechanism"* — partly conceded and now addressed within what the PR can do: the operator steps in the PR body are now unchecked task-list checkboxes (`AC-1`, `AC-5`, `AC-4`, each named), so the unmet criteria are a visible, unticked gate on the PR itself rather than prose. That is the strongest in-PR tracking available.

What remains genuinely open, and is yours to rule on: the PR body says `fixes #297`, so merging closes the spec issue while three of its ACs are still unticked. Two options:

- **Hold the merge** until you run the four steps on the GZ302 and tick the boxes — the ACs are then proven before #297 closes, and a step-3 failure (`claude: command not found`) is fixed in this PR instead of a follow-up.
- **Merge now** and keep the verification tracked separately — in which case tell me and I will file the on-device-verification follow-up issue (AC-1/AC-4/AC-5 + the fnm-PATH risk) before merge, or drop `fixes #297` to `refs #297` so the spec issue stays open until the demonstration lands.

I am not choosing between those on your behalf; both are one comment away once you rule.

roles/home/templates/hanzo-claude-code.desktop.j2
```diff
@@ -1,8 +1,19 @@
+{#
```
palazzem: This comment is unneeded.

roles/home/tasks/main.yml
```diff
@@ -73,6 +73,29 @@
 - name: Deploy Claude Code launcher .desktop entry
   ansible.builtin.template:
     src: hanzo-claude-code.desktop.j2
-    dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/hanzo-claude-code.desktop"
+    dest: "{{ ansible_facts.env.HOME }}/.local/share/applications/{{ home_claude_launcher_file }}"
     mode: "0644"
   become: false
+
+# kglobalacceld (Plasma 6) reads launcher shortcuts from the *nested*
```
palazzem: Simplify the comment. Explain the "why" in a simple task. Is it worth it to have a separate file for all KDE changes / updates? We're planning to make more (e.g. the UI configuration).

roles/home/templates/hanzo-claude-code.desktop.j2
palazzem: **Claude Harness**

Done — commit `7bd3a49`. The whole `{# ... -#}` block is gone; the template now starts at `[Desktop Entry]`.

The rendered file is byte-identical to before (the Jinja comment was stripped at render time anyway), confirmed by the container run:

```
TASK [home : Deploy Claude Code launcher .desktop entry] ***********************
+[Desktop Entry]
+Type=Application
+Name=Claude Code
+Comment=Open Claude Code in ~/cowork
+Exec=ghostty --working-directory=/home/testuser/cowork -e fish -c claude
+Icon=utilities-terminal
+Terminal=false
+Categories=Utility;Development;
```

The two facts the block carried (`%h` is not a Desktop Entry field code, and `claude` is only on PATH via fish because it is an fnm-managed npm global) stay recorded in the PR body's Proposed Changes section, so the reasoning is findable without carrying it in the template.

roles/home/tasks/main.yml
palazzem: **Claude Harness**

Done — commit `7bd3a49`. Both parts:

**Separate file: yes, worth it.** The KDE user-side tasks now live in `roles/home/tasks/kde.yml`, included from `main.yml` with a single task:

```yaml
- name: Configure KDE Plasma user settings
  ansible.builtin.include_tasks: kde.yml
```

It follows the repo's existing split convention (`hardware/gz302.yml`, `system/hibernate.yml`, `languages/*.yml`, `tools/*.yml`) and gives the planned UI configuration an obvious landing place instead of growing `main.yml`. Two deliberate choices so this stays a pure move:

- The include is **unconditional**; the `host_has_kde` gate stays on the binding task itself, exactly as before. Gating the include would also skip the launcher deployment on non-KDE hosts — a behavior change the issue does not ask for, and it would remove the container's AC-2 evidence (the container is non-KDE).
- `~/cowork` stays in `main.yml` — it is a plain workspace directory, not KDE config.

I did **not** move this into a new role or into `desktop`: role placement is a design call, and #297 records the binding as home-role work. If the coming KDE UI configuration makes `home` the wrong owner, that is worth deciding when that work is specced — say the word and I will raise it rather than drift it here.

**Comments simplified.** The 12-line derivation above the binding task is now three lines stating only the why, and the `home_claude_launcher_file` comment in `vars/main.yml` went from six lines to three:

```yaml
# kglobalacceld reads launcher shortcuts only from the nested
# `[services][<file>.desktop]` group; top-level `[<file>.desktop]` groups
# are ignored and deleted on daemon start. ini_file has no nested-group
# syntax, hence the `]`+`[` embedded in `section`.
```

The kglobalacceld 6.7.4 source derivation (`loadSettings()` / `migrateConfig()` / `KServiceActionComponent`) stays in the PR body's Proposed Changes section, where it is a review record rather than source noise.

Verified after the change:

- `pre-commit run --all-files` — all hooks pass (ansible-lint included).
- `docker build -f tests/Containerfile -t hanzo:test .` — `ok=61 changed=31 failed=0`; `Exec=ghostty --working-directory=/home/testuser/cowork -e fish -c claude` renders unchanged and the binding task still reports `skipping: [localhost]` (EC-1 intact).
- CI on `7bd3a49`: `lint` pass (1m8s), `test` pass (1m32s).

PR #302, opened 2026-08-17
fix(system): replace initramfs handler with convergent rebuild task (#293)

### LINKED ISSUE #293: Handler-driven initramfs rebuild is not convergent
**Claude Harness**

Filed as the tracking record for two review-panel findings on #291 (`reliability/F1`, `reliability/F2`) that are real but repo-wide, and whose remedy is a play-level / cross-role mechanism rather than part of the #290 hibernation bug fix.

## Problem

`roles/system/handlers/main.yml` (`Rebuild initramfs`) and `roles/hardware/handlers/main.yml` (`Regenerate initramfs`) both run `limine-mkinitcpio` as a handler, notified by tasks that write `/etc/default/limine` and `/etc/mkinitcpio.conf.d/`. Two ways the config and the generated boot artifacts can permanently diverge:

1. **Notification lost on a later failure.** Handlers flush at end of play. `playbook.yml` runs all eleven roles in one play, with ten roles (including network-dependent AUR builds) after `system`. Neither `ansible.cfg` nor `playbook.yml` sets `force_handlers`. If any later task fails, the config files are already written but `limine-mkinitcpio` never runs — and because the config tasks are idempotent, a rerun reports `ok`/`changed=0` and never re-notifies. The boot entries and initramfs stay stale forever while provisioning reports clean.

2. **Failed handler never retried.** If `limine-mkinitcpio` itself exits non-zero once (transient ESP-full, interrupted run, mid-transaction kernel state), the notification is consumed by that failing run. Subsequent runs change nothing, notify nothing, and the handler never re-runs. `force_handlers` does not cover this case.

Concretely for #290: `/etc/default/limine` can contain `systemd.zram=0` plus the `zswap.*` params and `/etc/mkinitcpio.conf.d/10-zswap.conf` can exist while the machine keeps booting with ZRam active and no `lz4` in the initramfs — i.e. hibernation still broken, with every `hanzo` run green.

## Directions (not decided)

- `force_handlers: true` on the play, or `meta: flush_handlers` after boot-config tasks — closes (1), not (2). Play-level error-handling semantics for all eleven roles, so it wants a deliberate decision.
- A convergence guard for (2): verify the generated Limine entries / initramfs actually carry the expected params and run the rebuild as a task with a real `changed_when`, instead of relying on notification state that a single failure consumes.

Refs: #290, #291

palazzem:
**Claude Harness**

Makes the initramfs / Limine rebuild convergent. The play-level `Regenerate initramfs` handler and its five `notify` lines are deleted; `limine-mkinitcpio` now runs unconditionally as the last task of the `system` role (tagged `always`, imported after `hardware`, gated on `host_has_systemd`). Handler notification state is consumed by a single run — lost when a later role fails before handlers flush, or spent by a rebuild that itself failed — and the idempotent config tasks never re-send it, so the boot artifacts could stay stale forever while every run reported clean. `force_handlers: true` is added on the play as hardening for the remaining service handlers (`Restart keyd`, `Reload logind`).

## Run record

- **Issue:** #293
- **Session:** `4a602ac4-4426-4869-93e7-67c017f63d2d` — resume with `claude --resume 4a602ac4-4426-4869-93e7-67c017f63d2d`

Closes #293

### DIFF
--- a/roles/system/tasks/main.yml
+++ b/roles/system/tasks/main.yml
@@ -24,12 +24,16 @@
   ansible.builtin.include_tasks: hibernate.yml
   when: host_has_systemd
 
-# Rebuilt unconditionally, last in the role (tagged `always`, imported
-# after `hardware`), because a handler notification is consumed by a single
-# run: lost or failed once, it is never re-sent and the boot artifacts stay
-# stale forever while every run reports clean (issue #293).
+# Runs on every invocation, last in the role, because a handler notification
+# is consumed once: lost or failed, it is never re-sent (issue #293).
 - name: Regenerate initramfs and Limine boot entries
   ansible.builtin.command: limine-mkinitcpio
+  register: system_limine_rebuild
   changed_when: true
+  # limine-mkinitcpio swallows per-kernel failures and still exits 0, printing
+  # `ERROR:` to stderr — rc alone would report a stale initramfs as success.
+  failed_when: >-
+    system_limine_rebuild.rc != 0
+    or 'ERROR:' in system_limine_rebuild.stderr
   when: host_has_systemd
   become: true
--- a/roles/system/tasks/main.yml
+++ b/roles/system/tasks/main.yml
@@ -23,3 +23,13 @@
 - name: Configure hibernation
   ansible.builtin.include_tasks: hibernate.yml
   when: host_has_systemd
+
+# Rebuilt unconditionally, last in the role (tagged `always`, imported
+# after `hardware`), because a handler notification is consumed by a single
+# run: lost or failed once, it is never re-sent and the boot artifacts stay
+# stale forever while every run reports clean (issue #293).
+- name: Regenerate initramfs and Limine boot entries
+  ansible.builtin.command: limine-mkinitcpio
+  changed_when: true
+  when: host_has_systemd
+  become: true
--- a/playbook.yml
+++ b/playbook.yml
@@ -8,6 +8,10 @@
     - "!all"
     - min
     - hardware
+  # Handlers flush at the end of the play, so without this a failure in any
+  # later role silently drops pending notifications that idempotent tasks
+  # will never re-send (issue #293).
+  force_handlers: true
 
   vars:
     # Transient sudoers drop-in granting NOPASSWD: /usr/bin/pacman
@@ -151,11 +155,3 @@
             state: absent
           become: true
           tags: [always]
-
-  # Regenerates the initramfs for every installed kernel and rewrites the
-  # Limine boot entries from /etc/default/limine.
-  handlers:
-    - name: Regenerate initramfs
-      ansible.builtin.command: limine-mkinitcpio
-      changed_when: true
-      become: true
--- a/playbook.yml
+++ b/playbook.yml
@@ -8,9 +8,8 @@
     - "!all"
     - min
     - hardware
-  # Handlers flush at the end of the play, so without this a failure in any
-  # later role silently drops pending notifications that idempotent tasks
-  # will never re-send (issue #293).
+  # Handlers flush at the end of the play, so a failure in a later role would
+  # otherwise drop pending notifications that idempotent tasks never re-send.
   force_handlers: true
 
   vars:
--- a/roles/hardware/tasks/gz302.yml
+++ b/roles/hardware/tasks/gz302.yml
@@ -47,7 +47,6 @@
     dest: /etc/modprobe.d/amdgpu.conf
     mode: "0644"
   become: true
-  notify: Regenerate initramfs
 
 - name: Ensure OLED kernel param in Limine cmdline
   ansible.builtin.lineinfile:
@@ -55,7 +54,6 @@
     regexp: '^KERNEL_CMDLINE\[default\]\+=" amdgpu\.dcdebugmask='
     line: 'KERNEL_CMDLINE[default]+=" {{ hardware_gz302_oled_kernel_param }}"'
   become: true
-  notify: Regenerate initramfs
 
 - name: Deploy suspend/resume hook
   ansible.builtin.template:
--- a/roles/system/tasks/hibernate.yml
+++ b/roles/system/tasks/hibernate.yml
@@ -7,15 +7,14 @@
 # https://wiki.cachyos.org/configuration/general_system_tweaks/#swap-configuration
 
 # Kernel params live in /etc/default/limine; limine-mkinitcpio regenerates
-# the boot entries and the initramfs from it, so both param tasks and the
-# mkinitcpio drop-in below share the same handler.
+# the boot entries and the initramfs from it — that rebuild runs
+# unconditionally at the end of the role (tasks/main.yml), not via handler.
 - name: Ensure ZRam is disabled in Limine cmdline
   ansible.builtin.lineinfile:
     path: /etc/default/limine
     regexp: '^KERNEL_CMDLINE\[default\]\+=" systemd\.zram='
     line: 'KERNEL_CMDLINE[default]+=" {{ system_zram_kernel_param }}"'
   become: true
-  notify: Regenerate initramfs
 
 - name: Ensure Zswap kernel params in Limine cmdline
   ansible.builtin.lineinfile:
@@ -23,7 +22,6 @@
     regexp: '^KERNEL_CMDLINE\[default\]\+=" zswap\.'
     line: 'KERNEL_CMDLINE[default]+=" {{ system_zswap_kernel_params }}"'
   become: true
-  notify: Regenerate initramfs
 
 # /usr/lib/udev/rules.d/30-zram.rules (cachyos-settings) writes `N` to
 # /sys/module/zswap/parameters/enabled whenever zram0 changes state. A
@@ -89,7 +87,6 @@
     group: root
     mode: "0644"
   become: true
-  notify: Regenerate initramfs
 
 # systemd ships sleep.conf and logind.conf, but not their .conf.d/
 # drop-in directories — ansible.builtin.copy does not create parent
--- a/roles/system/tasks/hibernate.yml
+++ b/roles/system/tasks/hibernate.yml
@@ -6,9 +6,8 @@
 # the CachyOS wiki "Switching from ZRam to Zswap" procedure:
 # https://wiki.cachyos.org/configuration/general_system_tweaks/#swap-configuration
 
-# Kernel params live in /etc/default/limine; limine-mkinitcpio regenerates
-# the boot entries and the initramfs from it — that rebuild runs
-# unconditionally at the end of the role (tasks/main.yml), not via handler.
+# Kernel params live in /etc/default/limine; the rebuild that applies them
+# runs at the end of the role (tasks/main.yml), not via handler.
 - name: Ensure ZRam is disabled in Limine cmdline
   ansible.builtin.lineinfile:
     path: /etc/default/limine


palazzem:
**Claude Harness**

Review-panel angle selection for this PR (deterministic rubric, one reviewer per angle):

**Core — always on:**
- `correctness` — the diff does what it claims: logic, boundary handling, error paths
- `divergence` — implementation vs issue #293 and its recorded design ruling (the **Claude Harness** comment on #293); every departure, omission, invention
- `testing-quality` — the tests prove the claims: ACs covered, assertions that can fail, no tautologies
- `simplicity` — abstraction, indirection, or generality no requirement demands
- `settling` — the diff extends a flawed convention where a better approach is evident

**Conditional:**
- `security` — **skipped** (no authn/authz, secret handling, crypto, or input parsing in the diff)
- `data` — **skipped** (no schema, migration, or query text)
- `concurrency` — **skipped** (no parallel execution, shared mutable state, locks, or async coordination)
- `reliability` — **fired**: the diff adds a subprocess call across a process boundary (`limine-mkinitcpio` task in `roles/system/tasks/main.yml`) and changes recovery logic (`force_handlers: true` in `playbook.yml`, handler deletion)
- `performance` — **skipped** (no loop-body changes, no caching)

**Repo-supplied:** none (`.claude/review-angles/` absent).

palazzem (commented):
**Claude Harness**

Examined the full diff — `playbook.yml` (handler deletion, `force_handlers: true`), `roles/hardware/tasks/gz302.yml` and `roles/system/tasks/hibernate.yml` (five `notify` removals, comment rewrite), `roles/system/tasks/main.yml` (new unconditional `limine-mkinitcpio` task) — for abstraction, indirection, or generality no requirement demands. The change is net-simplifying: it replaces the notify/handler indirection with one direct task. Every remaining element is demanded verbatim by the ruling on issue #293 (`force_handlers: true` as hardening for the surviving service handlers; the task placed last in the `system` role, gated `when: host_has_systemd`, with `changed_when: true`). Verified in the worktree that no orphaned indirection is left behind: all three surviving handlers (`Restart network stack`, `Restart keyd`, `Reload logind`) still have live notifiers, and no reference to the deleted `Regenerate initramfs` handler remains.

palazzem (commented):
**Claude Harness**

**Angle `settling` — run and clean**

Examined every convention this diff adopts or carries forward: the placement of the new `Regenerate initramfs and Limine boot entries` task at the end of `roles/system/tasks/main.yml` (role-ordering + `[system, always]` tag coupling vs the evident alternative of play-level `post_tasks` — the chosen placement is strictly more robust on the dominant failure axis, since it rebuilds immediately after both boot-config writer roles and before the failure-prone AUR-heavy roles, whereas `post_tasks` would leave boot artifacts stale for the whole aborted run, i.e. the exact #293 scenario); the `ansible.builtin.command` + `changed_when: true` shape (matches the repo's documented escape-hatch rule and the ruled unconditional-rebuild design); the reuse of the `host_has_systemd` detection fact from play `pre_tasks` (the established gating convention, not a redefinition); `force_handlers: true` as play-level hardening for the surviving service handlers (`Restart keyd`, `Reload logind` are untouched, not extended, and converting them to unconditional tasks would restart services on every run — worse); and the comment updates in `roles/system/tasks/hibernate.yml` and `playbook.yml` that retire the shared-handler description. Verified via `grep` that all `/etc/default/limine` and `mkinitcpio.conf.d` writers (`roles/hardware/tasks/gz302.yml`, `roles/system/tasks/hibernate.yml`) execute before the rebuild in every tag-selected invocation. No flawed convention is extended where a better approach is evident.

roles/system/tasks/main.yml
```diff
@@ -23,3 +23,13 @@
 - name: Configure hibernation
   ansible.builtin.include_tasks: hibernate.yml
   when: host_has_systemd
+
+# Rebuilt unconditionally, last in the role (tagged `always`, imported
+# after `hardware`), because a handler notification is consumed by a single
+# run: lost or failed once, it is never re-sent and the boot artifacts stay
+# stale forever while every run reports clean (issue #293).
+- name: Regenerate initramfs and Limine boot entries
```
palazzem, line 29: **Claude Harness**

**`testing-quality/F1` — The done-criterion's container check never executes this task: it reports `skipping` in the check build, so the green result contains no assertion that can fail on the PR's core change**

- **Evidence:** Ran both done-criterion checks on PR head `70043dc` in a fresh worktree. `pre-commit run --all-files` → all hooks Passed. `docker build --progress=plain -f tests/Containerfile -t hanzo:review302 .` → exit 0, with this in the ansible output:

  ```
  TASK [system : Configure hibernation] ******************************************
  skipping: [localhost]
  TASK [system : Regenerate initramfs and Limine boot entries] *******************
  skipping: [localhost]
  ...
  PLAY RECAP: localhost : ok=61 changed=31 failed=0 skipped=28
  ```

  `/run/systemd/system` is absent under `docker build` RUN (playbook.yml:26-30 documents exactly this), so `host_has_systemd` is false and the whole changed path is skipped: this task, the `hibernate.yml` tasks whose `notify` lines were removed, and the gz302 tasks (`model_is_gz302` also requires systemd). Even on a systemd host, `--check` mode skips `ansible.builtin.command`, so no variant of the check build ever reaches the command. `force_handlers: true` is equally unexercised — it only changes behavior on a task failure, and the check build contains none.
- **Impact:** The two done-criterion checks pass identically whether this task exists, is mis-gated, or `limine-mkinitcpio` is misspelled — the command string is never validated, so a typo here ships green. The PR's central claim ("the boot artifacts can no longer stay stale while every run reports clean") and issue #293's failure modes (1) and (2) are covered by zero assertions that can fail; the green CI on this PR proves only that the YAML parses and the untouched roles still check-run.

palazzem (commented):
**Claude Harness**

Examined whether the tests prove this PR's claims: ran the issue's done-criterion — `pre-commit run --all-files` and the container check build (`docker build -f tests/Containerfile …`) — against PR head `70043dc` in a fresh isolated worktree, then traced which changed lines those checks actually exercise: the new `Regenerate initramfs and Limine boot entries` task in `roles/system/tasks/main.yml`, the removed `notify` lines in `roles/system/tasks/hibernate.yml` and `roles/hardware/tasks/gz302.yml`, and `force_handlers: true` in `playbook.yml`. Both checks pass (CI is green on the same two jobs), but the build log proves the entire changed path is skipped.

- `testing-quality/F1` — the container check build skips the new rebuild task (and every rebuild-related config task), so the done-criterion cannot fail on this PR's core change → roles/system/tasks/main.yml:31

palazzem (commented):
**Claude Harness**

Examined the full diff (`playbook.yml`, `roles/system/tasks/main.yml`, `roles/system/tasks/hibernate.yml`, `roles/hardware/tasks/gz302.yml`) against issue #293 and its recorded design ruling (the **Claude Harness** comment on #293; no design-doc file exists). Every ruling item is implemented without departure, omission, or invention: the play-level `Regenerate initramfs` handler and all 5 `notify` lines are deleted (grep over `roles/` and `playbook.yml` finds zero remaining references); the unconditional `limine-mkinitcpio` task is the last task of the `system` role with `become: true`, `changed_when: true`, `when: host_has_systemd`, and the role is tagged `[system, always]` and imported after `hardware` (playbook.yml:96-104), so it runs on every invocation after both boot-config writer roles; `force_handlers: true` is set on the play; the now-false shared-handler comments (`hibernate.yml` header, playbook handler comment) are updated or removed. Done-criterion verified in an isolated worktree at the PR head (70043dc): `pre-commit run --all-files` passes and the container check build (`docker build -f tests/Containerfile`) completes with `failed=0`, with the new task reported `skipping` in the container — matching the ruling's stated expectation for the `host_has_systemd` gate.

palazzem (commented):
**Claude Harness**

Examined whether the diff does what it claims: handler-to-task conversion parity (all 5 `notify: Regenerate initramfs` lines present on `main` — 2 in `roles/hardware/tasks/gz302.yml`, 3 in `roles/system/tasks/hibernate.yml` — are removed along with the single play-level handler; `git grep` confirms no dangling notify remains), placement and ordering of the new rebuild task (last in `roles/system/tasks/main.yml`, after the `hibernate.yml` include; `system` imports after `hardware`, and a grep across all nine later roles finds no limine/mkinitcpio/modprobe writer, so relocating execution from end-of-play to end-of-`system` misses nothing), tag and gate behavior (`--list-tasks` in the CachyOS test container shows the task tagged `[always, system]` and still scheduled under a selective `--tags hardware` run; the gz302 writers are gated on `model_is_gz302`, which itself requires `hanzo_systemd_stat.stat.exists`, so no config writer can run where the rebuild's `when: host_has_systemd` gate is false), and error paths (a failed `limine-mkinitcpio` now fails the play and re-executes unconditionally on the next run — convergent by construction; `force_handlers: true` is a valid play keyword and the three remaining notified handlers — `Restart network stack`, `Restart keyd`, `Reload logind` — all still exist). The full container check build passed in an isolated clone of the branch (`docker build -f tests/Containerfile`: `ok=61 changed=31 failed=0`), with the rebuild task correctly reporting `skipping` in the non-systemd container, matching pre-PR behavior. No defect found.

roles/system/tasks/main.yml
```diff
@@ -23,3 +23,13 @@
...
+  ansible.builtin.command: limine-mkinitcpio
```
palazzem, line 30: **Claude Harness**

**`reliability/F1` — `limine-mkinitcpio` exits 0 when mkinitcpio fails for a kernel, so this task reports success (`rc=0`, `changed`) while the boot artifacts stay stale**

- **Evidence:** Live repro in a privileged `cachyos/cachyos:latest` container: installed `limine-mkinitcpio-hook 1.37.1-1` + `linux-cachyos`, loop-mounted a vfat ESP on `/boot`, replaced `/usr/bin/mkinitcpio` with a stub that exits 1, then ran `limine-mkinitcpio`:
  ```
  Building initramfs for linux-cachyos (7.1.8-1-cachyos)
  ==> ERROR: simulated mkinitcpio failure
  ERROR: mkinitcpio failed for kernel 7.1.8-1-cachyos, skipping.
  EXIT_CODE=0
  ```
  Traced cause in `/usr/share/libalpm/scripts/limine-mkinitcpio-install` (which `/usr/bin/limine-mkinitcpio` pipes `rebuild` into): `process_regular_kernel || return 0` inside `process_kernel`, `process_kernel "$pkgbase_file" || true` in `main`, and the `install_kernel` return value (the ESP write — the "transient ESP-full" case issue #293 names explicitly) is never checked. Every per-kernel failure is swallowed; only source/ESP-detection failures (exit 1) and boot hooks with rc >= 100 (exit 2/3) propagate. `ansible.builtin.command` fails only on non-zero rc, and `changed_when: true` marks the task changed unconditionally.
- **Impact:** The run in which the rebuild fails reports green and `changed`, so the operator can reboot or hibernate into a stale initramfs (e.g. no `lz4` hook — #290's exact symptom) with zero signal. With a persistent cause (ESP full, broken mkinitcpio config) every subsequent run also exits 0: the "boot artifacts stay stale forever while every run reports clean" state that #293 exists to close is reopened inside the subprocess. Registering the result and failing on the script's stderr `ERROR:` marker (`error_msg` writes to stderr) would surface it.

palazzem (commented):
**Claude Harness**

**Angle `reliability` — 2 finding(s)**

Examined the added unconditional `limine-mkinitcpio` task (`roles/system/tasks/main.yml:31-35`) — exit-code semantics, partial-failure behavior, timeout exposure, recovery guarantees — and the changed handler-recovery semantics: deletion of the play-level `Regenerate initramfs` handler and its five `notify` lines (`playbook.yml`, `roles/hardware/tasks/gz302.yml`, `roles/system/tasks/hibernate.yml`) plus the added `force_handlers: true`. Verified the shipped `limine-mkinitcpio-hook 1.37.1-1` scripts and ran a live failure repro in a privileged `cachyos/cachyos:latest` container; verified tag-selection semantics with in-container `--list-tasks`. Checked clean: timeout exposure (the boot-partition mutex uses `flock --timeout=10`, no unbounded-wait path added) and `force_handlers: true`, which correctly hardens the two surviving service handlers (`Restart keyd`, `Reload logind`) against the lost-notification mode.

- `reliability/F1 — limine-mkinitcpio exits 0 when mkinitcpio fails, so the task reports success while boot artifacts stay stale → roles/system/tasks/main.yml:32`
- `reliability/F2 — the rebuild task is tag-filterable while the deleted handler was not; hardware's limine writers can run without any rebuild → roles/system/tasks/main.yml:31`

roles/system/tasks/main.yml
palazzem, line 29: **Claude Harness**

**`reliability/F2` — the rebuild is now tag-filterable while the handler it replaces was not: an invocation that runs `hardware`'s limine writers but filters this task writes boot config with no rebuild**

- **Evidence:** Ansible docs (`playbook_guide/playbooks_tags.rst`, devel): "Handlers are a special type of task that only execute when explicitly notified. They ignore all tags and cannot be selected for or against using tag-based filtering." — the deleted play-level `Regenerate initramfs` handler therefore ran under any tag selection once notified. The replacement task inherits `[system, always]` and is filterable: in-container `ansible-playbook playbook.yml --list-tasks --skip-tags system` lists `hardware : Include GZ302 hardware fixes	TAGS: [hardware]` (containing the `/etc/default/limine` and `amdgpu.conf` writers) but no `Regenerate initramfs and Limine boot entries` task. `bin/hanzo` forwards arbitrary `ansible-playbook` args (`ansible-playbook playbook.yml --ask-become-pass "$@"`), so the invocation is reachable.
- **Impact:** `hanzo --skip-tags system` on a GZ302 host updates the Limine cmdline and modprobe config, exits green, and never rebuilds — boot entries stay stale until the next run that includes `system`, a window the notify mechanism did not have. Narrow (undocumented invocation, self-healing on the next full run), but a strict recovery-coverage regression introduced by deleting the handler.

roles/system/tasks/main.yml
palazzem: **Claude Harness**

**Fixed** in `5a15e8cccbf38ee6ffd00f4784fb67041c657f39`.

Confirmed the mechanism independently in `cachyos/cachyos:latest` with `limine-mkinitcpio-hook` installed:

- `/usr/lib/limine/limine-common-functions:30` — `error_msg() { echo -e "${_color_red}ERROR: $1${_color_reset} ${2:-}" >&2; }`, i.e. every swallowed failure is reported only on stderr.
- `/usr/share/libalpm/scripts/limine-mkinitcpio-install` — `process_regular_kernel || return 0` (line 250), `process_kernel "$pkgbase_file" || true` (line 265), and the three `error_msg "mkinitcpio failed for kernel ..."` sites at lines 183/212/225. Exit stays 0.

The task now registers the result and fails on rc **or** the stderr marker:

```yaml
  register: system_limine_rebuild
  changed_when: true
  failed_when: >-
    system_limine_rebuild.rc != 0
    or 'ERROR:' in system_limine_rebuild.stderr
```

Verified with a stubbed `limine-mkinitcpio` on the real task body inside the test container (never on the host):

```
### CASE A: exits 0 but printed ERROR: on stderr
fatal: [localhost]: FAILED! => {"failed_when_result": true, "rc": 0,
  "msg": "Task failed: Action failed: A 'failed_when' expression evaluated to 'True'.",
  "stderr": "[91mERROR: mkinitcpio failed for kernel 7.1.8-1-cachyos, skipping.[0m "}
localhost : ok=0 changed=0 failed=1

### CASE B: clean run, `==> WARNING:` on stderr
changed: [localhost]
localhost : ok=1 changed=1 failed=0
```

Case B shows mkinitcpio's normal `==> WARNING:` stderr output does not trip the guard; the marker is matched raw so the leading ANSI colour escape does not defeat it.

Repo checks on the new head: `pre-commit run --all-files` → all hooks Passed; `docker build -f tests/Containerfile -t hanzo:test .` → exit 0, `PLAY RECAP: ok=61 changed=31 failed=0 skipped=28`.

roles/system/tasks/main.yml
palazzem: **Claude Harness**

**Deferred** — the finding is accurate, but closing it is test-infrastructure work outside this change's scope. Tracking issue: #304.

Reproduced the observation and agree with it: on this branch the check build reports `skipping: [localhost]` for `Regenerate initramfs and Limine boot entries`, and `--check` would skip `ansible.builtin.command` even on a systemd host.

Two reasons it is not fixed here:

1. **It is not a regression of this PR.** The gap covers every systemd-gated task in the repo (`hibernate.yml`, `gz302.yml`), and the play-level `Regenerate initramfs` handler this PR deletes was equally unexercised — it was notified only by systemd-gated tasks and would have been skipped in `--check` as well. Before and after the diff, the container test asserts the same nothing about this path.
2. **Closing it requires a design decision the #293 record does not cover.** Executing the path needs a systemd-capable, non-`--check` test lane (privileged container with `/sbin/init`, `systemd-nspawn`, or a VM job) plus a decision on how to contain real root-level boot writes (`limine-mkinitcpio`, `/etc/default/limine`, the ESP). The repo's current test definition — the done-criterion recorded on #293 — is `pre-commit` plus the `--check --diff` container build; changing that definition is a fresh design pass, not builder execution.

What this PR does add, in place of harness coverage, is an assertion inside the runtime path itself: `reliability/F1`'s finding is fixed in `5a15e8c`, so a rebuild that fails now fails the run instead of reporting green. The command string is still only validated at real runtime; #304 records the follow-up, including a check that the new `failed_when` fires.

palazzem: **Claude Harness**

**Deferred** — the finding is factually correct and I am not contesting it, but the fix requires a design decision outside this run's bounds. Tracking issue: #305.

Confirmed: handlers ignore tag filtering, the replacement task does not, so `hanzo --skip-tags system` on a GZ302 host writes `/etc/default/limine` and `/etc/modprobe.d/amdgpu.conf` via the `hardware` role and never rebuilds. That window did not exist under the handler.

Why it is not fixed here: the task's placement — one unconditional task at the **end of the `system` role** — is the recorded design ruling on #293 (the **Claude Harness** comment there). Every available fix changes that placement or the tag scheme (play-level `post_tasks`, a dedicated `initramfs` tag, or a guard that fails when a boot-config writer runs without the rebuild). Choosing among those is a design decision the record does not cover, so I am not making it silently inside an execution run.

Scope note, consistent with the finding's own severity read: the invocation is undocumented, `--skip-tags system` also skips the hibernate writers it pairs with, and the state self-heals on the next run that includes `system`. #293's failure modes (1) and (2) — a notification lost to a later role's failure, or spent by a rebuild that itself failed, on an ordinary full run — are closed by this PR regardless. #305 carries the residual selective-run window.

PR #309, opened 2026-08-18
fix(cli): replace destructive checkout sync with a non-destructive guard

palazzem:
## What changed

`bin/hanzo` ran an unconditional `git fetch origin main && git reset --hard origin/main` before every provisioning run, destroying any local work in the checkout. That sync block is now a guard:

- No `origin` remote: warn and provision the working tree as-is.
- Dirty working tree: warn and skip the sync to keep local changes.
- Checkout not on `main`: warn and skip.
- Otherwise fetch, and fast-forward only (`git merge --ff-only`) when HEAD is an ancestor of `origin/main`; local commits not on `origin/main` are kept with a warning.

Adds a `log_warn` helper (yellow) alongside `log_info`. Everything else in the script (sudoers cleanup trap, collection refresh, playbook invocation) is unchanged.

## Why

The sync exists so machines provision the published configuration — it must never destroy local work.

## Test evidence

- `pre-commit run --files bin/hanzo`: all hooks passed (shellcheck included).
- `docker build -f tests/Containerfile -t hanzo:split-sync .`: exit 0; check run finished with `ok=61 changed=31 unreachable=0 failed=0`.

### DIFF
--- a/bin/hanzo
+++ b/bin/hanzo
@@ -8,8 +8,10 @@
 set -euo pipefail
 
 GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
 NC='\033[0m'
 log_info() { echo -e "${GREEN}[hanzo]${NC} $1"; }
+log_warn() { echo -e "${YELLOW}[hanzo]${NC} $1"; }
 
 HANZO_DIR="$(cd "$(dirname "$(readlink -f "$0")")/.." && pwd)"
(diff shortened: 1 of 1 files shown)


palazzem:
Closing as it is unneeded. There must never be intermediate changes. If you make changes, you commit. If you don't, you lose data. This is a provisioner from scratch, there is no reasons to have intermediate states.

PR #310, opened 2026-08-18
feat(security)!: drop proton-vpn-gtk-app in favor of manual WireGuard

palazzem:
## What changed

- `roles/security/vars/main.yml`: removed `proton-vpn-gtk-app`; the list became empty, so `security_pacman` is gone entirely. Header comment now states that VPN uses WireGuard directly, configured manually post-install.
- `roles/security/tasks/main.yml`: removed the now-orphaned pacman install task. The AUR task (1Password) is untouched.
- `bin/bootstrap.sh`: left unchanged — on main it ends with `exec hanzo`, so there is no post-provisioning point where a WireGuard setup pointer could print without altering the exec semantics.

## Why

The Proton VPN GTK client is no longer wanted; VPN connectivity is WireGuard configured manually after install, so the provisioner declares no VPN client package.

## Test evidence

- `pre-commit run --files roles/security/vars/main.yml roles/security/tasks/main.yml`: all hooks passed (ansible-lint included).
- `docker build -f tests/Containerfile -t hanzo:split-vpn .`: exit 0 — PLAY RECAP `ok=60 changed=30 failed=0`, image exported.
- `grep -rn 'security_pacman\|proton' roles/ playbook.yml`: no matches remain.

### DIFF
--- a/roles/security/vars/main.yml
+++ b/roles/security/vars/main.yml
@@ -1,9 +1,8 @@
 ---
-# Security tooling. VPN uses WireGuard directly, configured manually
-# post-install — no VPN client package. 1Password (GUI + CLI) is
-# AUR-only and depends on the AgileBits GPG key seeded by roles/trust —
-# playbook.yml orders trust before security, and trust is tagged
-# `always` so any --tags security run still imports the key first.
+# Security tooling. 1Password (GUI + CLI) is AUR-only and depends on
+# the AgileBits GPG key seeded by roles/trust — playbook.yml orders
+# trust before security, and trust is tagged `always` so any --tags
+# security run still imports the key first.
 security_aur:
   - 1password
   - 1password-cli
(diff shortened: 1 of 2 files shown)


roles/security/vars/main.yml
```diff
@@ -1,12 +1,9 @@
 ---
-# Security tooling. Proton VPN ships via pacman; 1Password (GUI + CLI)
-# is AUR-only and depends on the AgileBits GPG key seeded by
-# roles/trust — playbook.yml orders trust before security, and trust is
-# tagged `always` so any --tags security run still imports the key
-# first.
-security_pacman:
-  - proton-vpn-gtk-app
-
+# Security tooling. VPN uses WireGuard directly, configured manually
```
palazzem: Remove this comment that is unneeded. Do not explain WHY a package is not there, because it means you should explain why the entire pacman repository is not installed, right?

palazzem: Removed in 59fe2d7 — the vars header now documents only the 1Password/AgileBits key mechanics, nothing about absent packages.

PR #313, opened 2026-08-18
test(ci): provision the container through bootstrap.sh

palazzem:
Container tests now provision through `bin/bootstrap.sh` exactly as a production machine would, so CI exercises the real setup path instead of a hand-rolled ansible-playbook invocation.

- `bin/bootstrap.sh`: optional mode argument — `--check` (dry run), `--ci` (unattended full provisioning, never prompts), no argument keeps the attended flow; unknown arguments error out.
- `bin/hanzo`: passes `--ask-become-pass` only when a tty exists, and `HANZO_SKIP_SYNC=1` skips the origin/main sync so container tests run the code under review.
- `tests/Containerfile`: pre-copies the repo to bootstrap's clone path, replaces a worktree `.git` pointer with `git init -q`, and runs `bash bin/bootstrap.sh $HANZO_TEST_ARGS` (default `--check`); drops the duplicated uv/ansible/galaxy/config steps and paru.
- `CLAUDE.md`: Commands section updated to the `HANZO_TEST_ARGS` forms.

Tests: `pre-commit run --all-files` passes (all hooks); `docker build -f tests/Containerfile -t hanzo:ci-check .` exits 0 with `RUN bash bin/bootstrap.sh --check` completing the full bootstrap → hanzo → `ansible-playbook --check` path.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -15,11 +15,11 @@ CachyOS system provisioner powered by Ansible.
 
 All provisioning operations run inside the CachyOS test container — running on the host modifies system state (see Security: Prohibited Commands below).
 
+The container provisions through `bin/bootstrap.sh`, exactly as a production machine would.
+
 - `pre-commit run --all-files`: Lint all files (ansible-lint, shellcheck, generic hooks).
-- `docker build -f tests/Containerfile -t hanzo:test .`: Run `ansible-playbook --check --diff` (full provisioning check).
-- `docker build --build-arg ANSIBLE_ARGS="--list-tags" -f tests/Containerfile -t hanzo:test .`: List all available tags.
-- `docker build --build-arg ANSIBLE_ARGS="--tags <role> --check --diff" -f tests/Containerfile -t hanzo:test .`: Check a single tagged role.
-- `docker build --build-arg ANSIBLE_ARGS="" -f tests/Containerfile -t hanzo:test .`: Real provisioning run inside the container (no `--check`).
+- `docker build -f tests/Containerfile -t hanzo:test .`: Provisioning check (`bootstrap.sh --check`, the default and what CI runs).
+- `docker build --build-arg HANZO_TEST_ARGS="--ci" -f tests/Containerfile -t hanzo:test .`: Full unattended provisioning inside the container.
 
 ## Role Tags
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,9 +17,11 @@ All provisioning operations run inside the CachyOS test container — running on
 
 The container provisions through `bin/bootstrap.sh`, exactly as a production machine would.
 
+`HANZO_ARGS` has no default — every build passes it explicitly.
+
 - `pre-commit run --all-files`: Lint all files (ansible-lint, shellcheck, generic hooks).
-- `docker build -f tests/Containerfile -t hanzo:test .`: Provisioning check (`bootstrap.sh --check`, the default and what CI runs).
-- `docker build --build-arg HANZO_TEST_ARGS="--ci" -f tests/Containerfile -t hanzo:test .`: Full unattended provisioning inside the container.
+- `docker build --build-arg HANZO_ARGS="--check" -f tests/Containerfile -t hanzo:test .`: Provisioning check (dry run, what CI runs).
+- `docker build --build-arg HANZO_ARGS="--ci" -f tests/Containerfile -t hanzo:test .`: Full unattended provisioning.
 
 ## Role Tags
 
--- a/bin/bootstrap.sh
+++ b/bin/bootstrap.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Hanzo bootstrap — one-command CachyOS provisioner setup.
 # Usage: curl -L https://raw.githubusercontent.com/palazzem/hanzo/main/bin/bootstrap.sh | bash
+# Modes: --check (dry run), --ci (unattended full provisioning, CI only).
 
 set -euo pipefail
 
@@ -16,6 +17,25 @@ NC='\033[0m'
 log_info()    { echo -e "${GREEN}[hanzo]${NC} $1"; }
 log_warn()    { echo -e "${YELLOW}[hanzo]${NC} $1"; }
 log_error()   { echo -e "${RED}[hanzo]${NC} $1" >&2; }
+
+# ---------------------------------------------------------------------------
+# Mode
+# ---------------------------------------------------------------------------
+HANZO_MODE="${1:-}"
+
+if [ "$#" -gt 1 ]; then
+    log_error "Usage: bootstrap.sh [--check|--ci]"
+    exit 1
+fi
+
+case "$HANZO_MODE" in
+    ""|--check|--ci) ;;
+    *)
+        log_error "Unknown argument: $HANZO_MODE"
+        log_error "Usage: bootstrap.sh [--check|--ci]"
+        exit 1
+        ;;
+esac
 
 # ---------------------------------------------------------------------------
 # Banner
@@ -113,6 +133,10 @@ elif [ -n "${HANZO_FULLNAME:-}" ] && [ -n "${HANZO_EMAIL:-}" ]; then
 
     write_config
     log_info "Configuration saved to $CONFIG_FILE (from environment)"
+elif [ "$HANZO_MODE" = "--ci" ]; then
+    # Unattended mode must never prompt.
+    log_error "--ci requires HANZO_FULLNAME and HANZO_EMAIL (or an existing config)"
+    exit 1
 else
     log_info "First-time setup — configuring Hanzo"
     echo ""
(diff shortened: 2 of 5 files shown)


bin/hanzo
```diff
@@ -30,12 +30,25 @@ trap cleanup EXIT INT TERM
 
 cd "$HANZO_DIR"
 
-log_info "Syncing checkout to origin/main..."
-git fetch origin main
-git reset --hard origin/main
+# Container tests set HANZO_SKIP_SYNC=1 so the code under review runs
```
palazzem: We should not use arguments for the bootstrap and env variables for the script. Or we rely only on env vars, or on arguments. My take is that as `--check` is mandatory for Ansible, we should use arguments on all scripts: we defined it in the bootstrap and we propagate to all scripts below. I would recommend:
- Decompose Bash so that you have something to import with utility functions. I want to keep it simple, we don't need to implement a full CLI, just a simple `--check` and `--ci` argument parsing.
- You reuse such utilities in all scripts

bin/hanzo
palazzem, line 7: As we're about to decompose bash scripts, you can move everything that makes sense to these utility functions. Search for best practices and extend the scope of the change.

bin/bootstrap.sh
palazzem: As we're defining arguments like `--check` and `--ci`, I would write YELLOW messages (as warnings) to mention that we're in Check mode and CI mode. Design these messages in a way that are visually nice and that can scale if we add more flags.

palazzem: Is there a way to identify from bootstrap if we're in a container? I would like to exit and write an error message (red) that if you are not in a container and are using `--ci`, that is forbidden. This is a guardrail as it will skip AUR verification and will install all packages without having a human that checks it.

bin/hanzo
```diff
@@ -30,12 +30,25 @@ trap cleanup EXIT INT TERM
...
+# instead of origin/main; real machines always get the destructive sync.
+if [[ "${HANZO_SKIP_SYNC:-}" == "1" ]]; then
+    log_info "Skipping checkout sync (HANZO_SKIP_SYNC=1)..."
```
palazzem: you should log that this is running in CI mode, so we skip the checkout on main.

tests/Containerfile
```diff
@@ -1,33 +1,31 @@
...
 
-# ansible-playbook flag set. Default `--check --diff` validates the playbook
-# without modifying the system; override to `""` (or another flag set) to
-# run a real provisioning inside the container:
-#     docker build --build-arg ANSIBLE_ARGS="" -f tests/Containerfile -t hanzo:test .
-ARG ANSIBLE_ARGS="--check --diff"
+# Bootstrap mode. Default `--check` validates the playbook without modifying
+# the system; `--ci` runs a full unattended provisioning:
+#     docker build --build-arg HANZO_TEST_ARGS="--ci" -f tests/Containerfile -t hanzo:test .
+ARG HANZO_TEST_ARGS="--check"
```
palazzem: avoid defaults. Make Github actions explicit on what arguments they pass so that if they don't the build will fail. Also, call this `HANZO_ARGS`.

tests/Containerfile
```diff
@@ -1,33 +1,31 @@
...
-# only fire when the config is present).
-RUN mkdir -p /home/testuser/.config/hanzo && \
-    printf 'hanzo_fullname: "Test User"\nhanzo_email: "test@example.com"\n' \
-      > /home/testuser/.config/hanzo/config.yml
+# A worktree checkout ships a `.git` pointer file; make `.git` a real
+# directory so bootstrap detects an existing checkout.
+RUN if [ ! -d .git ]; then rm -f .git && git init -q; fi
+
+ENV HANZO_SKIP_SYNC=1
+ENV HANZO_FULLNAME="Test User"
```
palazzem: Move these at the top as they would not change.

bin/hanzo
```diff
@@ -30,12 +30,25 @@ trap cleanup EXIT INT TERM
...
+fi
 
 log_info "Refreshing Ansible collections..."
 ansible-galaxy collection install -r requirements.yml
 
+# Ask for the become password only when a terminal exists — container
+# builds have no tty and sudo is passwordless there.
+HANZO_BECOME_ARGS=()
+if { : </dev/tty; } 2>/dev/null; then
+    HANZO_BECOME_ARGS+=(--ask-become-pass)
```
palazzem: Call it `HANZO_ASK_PASS`. Also should that happen only if we're not in CI mode? If we're in CI, we should always avoid to ask pass.

CLAUDE.md
```diff
@@ -15,11 +15,11 @@ CachyOS system provisioner powered by Ansible.
...
 
+The container provisions through `bin/bootstrap.sh`, exactly as a production machine would.
+
 - `pre-commit run --all-files`: Lint all files (ansible-lint, shellcheck, generic hooks).
-- `docker build -f tests/Containerfile -t hanzo:test .`: Run `ansible-playbook --check --diff` (full provisioning check).
-- `docker build --build-arg ANSIBLE_ARGS="--list-tags" -f tests/Containerfile -t hanzo:test .`: List all available tags.
-- `docker build --build-arg ANSIBLE_ARGS="--tags <role> --check --diff" -f tests/Containerfile -t hanzo:test .`: Check a single tagged role.
-- `docker build --build-arg ANSIBLE_ARGS="" -f tests/Containerfile -t hanzo:test .`: Real provisioning run inside the container (no `--check`).
+- `docker build -f tests/Containerfile -t hanzo:test .`: Provisioning check (`bootstrap.sh --check`, the default and what CI runs).
+- `docker build --build-arg HANZO_TEST_ARGS="--ci" -f tests/Containerfile -t hanzo:test .`: Full unattended provisioning inside the container.
```
palazzem: Write the full unattended provisioning only.

palazzem (commented):
I left many comments. I would say to split this work in multiple PRs to make the code easy to review. Identify changes and group in small PRs. Then we'll finalize this PR.

palazzem:
**Claude Harness** Split plan: PR A — shared bash lib (logging, tty/container detection, --check/--ci parsing) adopted by bootstrap and hanzo; PR B (stacked on A) — mode flags end-to-end: args-only propagation, yellow mode banners, red --ci-outside-container guardrail, CI sync-skip log, ask-pass gated on mode+tty; this PR rebases last to Containerfile (HANZO_ARGS, no defaults, stable layers first) + explicit workflow args + CLAUDE.md.

bin/hanzo
palazzem: **Claude Harness** Shipped via #315: --check/--ci are arguments end to end — bootstrap validates them and propagates to hanzo through exec. The shared utility lib was built, then removed per your greenfield ruling; each script is now self-contained.

bin/hanzo
palazzem: **Claude Harness** Superseded — the utility lib landed as #314 and was then deleted in the #315 greenfield rewrite you approved; the two scripts are short enough that sharing cost more than it saved.

bin/bootstrap.sh
palazzem: **Claude Harness** Superseded by the approved #315 simplification — the banner machinery was removed; the --ci container refusal and ansible's own --check output carry the signal.

palazzem: **Claude Harness** Shipped via #315: bootstrap refuses --ci outside a container (red error, checks /run/.containerenv and /.dockerenv — verified present in docker build RUN steps).

bin/hanzo
palazzem: **Claude Harness** Superseded in abab164 — sync now runs only on a real attended run; --check and --ci never touch the checkout (in CI the reset would silently replace the code under review with main).

tests/Containerfile
palazzem: **Claude Harness** Done in abab164 — ARG HANZO_ARGS with no default, a guard layer fails the build when it is missing (verified), and the workflow passes --check explicitly.

tests/Containerfile
palazzem: **Claude Harness** Done in abab164 — packages, user, PATH and identity env sit above the mode-dependent layers.

bin/hanzo
palazzem: **Claude Harness** Shipped via #315's final form — the prompt flag joins the single ARGS array only when /dev/tty is reachable and the mode is not --ci.

CLAUDE.md
palazzem: **Claude Harness** Done in abab164 — the command reads "Full unattended provisioning."

bin/bootstrap.sh
palazzem: **Claude Harness** Correction: my /.dockerenv claim was wrong — the probe matched BuildKit's echoed command text, not the file. docker build RUN steps mount neither /.dockerenv nor /run/.containerenv, so the merged guard blocked --ci in the container. Fixed on #312 (4b340a6): tests/Containerfile sets container=docker and the guard accepts a non-empty $container.

PR #319, opened 2026-08-18
chore(claude): expand deny-list and add PreToolUse provisioning guard

palazzem:
## What changed

**`.claude/settings.json`**

- Added the missing deny entries for `hanzo-aur` in every form `hanzo` already had: bare, `bin/`, `./bin/`, each with the ` *` argument variant.
- Added interpreter forms for both entry points: `bash bin/hanzo`, `sh bin/hanzo`, `bash bin/hanzo-aur`, `sh bin/hanzo-aur` (+ ` *` variants). `bootstrap.sh` already had these.
- Added tilde forms: `~/.local/bin/hanzo`, `~/.local/bin/hanzo-aur` (+ ` *` variants).
- Added `ansible-pull` and `ansible-console` (+ ` *` variants). `ansible-galaxy` is deliberately **not** denied — it is harmless and needed for `requirements.yml`.
- Added `shelly install` and `shelly install *`. Read-only `shelly` subcommands stay allowed.
- Added a `hooks.PreToolUse` block matching `Bash`, running `$CLAUDE_PROJECT_DIR/.claude/hooks/guard-provisioning.sh`.

**`.claude/hooks/guard-provisioning.sh`** (new, mode 0755)

Reads the tool JSON on stdin, extracts `.tool_input.command` with `jq` (falling back to `python3 -c json` when `jq` is absent), and exits `2` with a one-line stderr explanation when the command *executes* a provisioning entry point.

The match is anchored to **execution positions** — start of line, or after `;` `|` `&` `&&` `||` `$(` `` ` `` / newline — optionally preceded by env assignments (`FOO=1`) and by pass-through wrappers (`env`, `sudo`, `command`, `nohup`, `bash`, `sh`, `zsh`, `source`, `.`). Targets: any path ending in `/hanzo`, `/hanzo-aur`, `/bootstrap.sh`; bare `hanzo` / `hanzo-aur`; `ansible`, `ansible-playbook`, `ansible-pull`, `ansible-console`; and `shelly install`.

## Why

The deny-list only blocked a handful of literal spellings, so whole families slipped past: interpreter invocations, `~/.local/bin` paths, and `ansible-pull`. Worse, `bin/hanzo-aur` and `shelly install` — which install AUR packages on the host — were not covered at all. The literal list is inherently spelling-based, so the hook backs it with positional parsing of the full command line, catching pipelines, subshells and env prefixes while leaving commands that merely name those paths as data untouched.

## Verification

`pre-commit run --all-files`:

```
check for added large files..............................................Passed
check for case conflicts.................................................Passed
check that executables have shebangs.....................................Passed
check json...............................................................Passed
check for merge conflicts................................................Passed
check that scripts with shebangs are executable..........................Passed
check for broken symlinks............................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
check vcs permalinks.....................................................Passed
check yaml...............................................................Passed
detect destroyed symlinks................................................Passed
detect private key.......................................................Passed
fix end of files.........................................................Passed
forbid new submodules................................(no files to check)Skipped
mixed line ending........................................................Passed
trim trailing whitespace.................................................Passed
```

Hook test matrix — each command was fed to the hook as `{"tool_name":"Bash","tool_input":{"command":"..."}}` on stdin and the exit code asserted:

```
=== MUST BLOCK (exit 2) ===
PASS  exit=2  ansible-playbook playbook.yml
PASS  exit=2  cd /tmp && ansible-playbook x.yml
PASS  exit=2  bash bin/hanzo
PASS  exit=2  bash /Users/x/hanzo/bin/bootstrap.sh --check
PASS  exit=2  ~/.local/bin/hanzo-aur
PASS  exit=2  echo y | shelly install aur foo
PASS  exit=2  ansible-pull -U repo
PASS  exit=2  ./bin/hanzo --check
PASS  exit=2  env hanzo
PASS  exit=2  source bin/bootstrap.sh
PASS  exit=2  FOO=1 hanzo

=== MUST ALLOW (exit 0) ===
PASS  exit=0  cat bin/hanzo
PASS  exit=0  git add bin/hanzo-aur
PASS  exit=0  shellcheck bin/bootstrap.sh
PASS  exit=0  docker build --build-arg HANZO_ARGS="--check" -f tests/Containerfile -t hanzo:test .
PASS  exit=0  grep -rn ansible roles/
PASS  exit=0  git log -- bin/hanzo
PASS  exit=0  shelly --help
PASS  exit=0  pre-commit run --all-files
PASS  exit=0  echo hanzo

ALL TESTS PASSED
```

The `jq`-less path was exercised separately by running the hook with `PATH=/usr/bin:/bin` (no `jq` on that path); it still blocked `ansible-playbook playbook.yml` with exit 2.

No container build was run — this PR touches no Ansible or shell provisioning code, only harness configuration, so `pre-commit` is the applicable gate.

## Caveats

- The guard is a heuristic over the raw command string, not a shell parser. A provisioning command hidden inside a nested quoted string (e.g. `bash -c "hanzo"`) or reached through an alias or a wrapper script is not detected. It is a second layer behind the deny-list and behind `CLAUDE.md`, not a sandbox.
- `grep` matches per line, so a real newline in a command is treated as a separator — which is the intended behaviour.
- `ansible-galaxy` and read-only `shelly` subcommands remain allowed by design.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -5,12 +5,16 @@
       "Bash(ansible *)",
       "Bash(ansible-playbook)",
       "Bash(ansible-playbook *)",
+      "Bash(ansible-pull)",
+      "Bash(ansible-pull *)",
+      "Bash(ansible-console)",
+      "Bash(ansible-console *)",
       "Bash(hanzo)",
       "Bash(hanzo *)",
-      "Bash(./bin/hanzo)",
-      "Bash(./bin/hanzo *)",
-      "Bash(bin/hanzo)",
-      "Bash(bin/hanzo *)",
+      "Bash(hanzo-aur)",
+      "Bash(hanzo-aur *)",
+      "Bash(shelly install)",
+      "Bash(shelly install *)",
       "Bash(./bin/bootstrap.sh)",
       "Bash(./bin/bootstrap.sh *)",
       "Bash(bin/bootstrap.sh)",
@@ -19,6 +23,19 @@
       "Bash(bash bin/bootstrap.sh *)",
       "Bash(sh bin/bootstrap.sh)",
       "Bash(sh bin/bootstrap.sh *)"
+    ]
+  },
+  "hooks": {
+    "PreToolUse": [
+      {
+        "matcher": "Bash",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "$CLAUDE_PROJECT_DIR/.claude/hooks/guard-provisioning.sh"
+          }
+        ]
+      }
     ]
   }
 }
--- a/.claude/hooks/guard-provisioning.sh
+++ b/.claude/hooks/guard-provisioning.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+# PreToolUse guard: block any Bash command that executes a provisioning entry
+# point on the host. Commands that only reference those paths as data (cat,
+# grep, git, shellcheck, docker build) are allowed through.
+set -euo pipefail
+
+PAYLOAD=$(cat)
+
+extract_command() {
+    if command -v jq >/dev/null 2>&1; then
+        printf '%s' "$PAYLOAD" | jq -r '.tool_input.command // ""'
+    else
+        printf '%s' "$PAYLOAD" | python3 -c 'import json,sys; print(json.load(sys.stdin).get("tool_input", {}).get("command", ""))'
+    fi
+}
+
+COMMAND=$(extract_command 2>/dev/null || printf '')
+[ -n "$COMMAND" ] || exit 0
+
+# An execution position is the start of a line or the character after a shell
+# separator (; | & && || $( `), optionally preceded by env assignments and by
+# wrappers or interpreters that pass through to the real command.
+SEPARATOR='(^|[;&|(`])[;&|]*[[:space:]]*'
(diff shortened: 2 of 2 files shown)


.claude/settings.json
palazzem, line 4: I think all these denies are extreme. Can't you aggregate it in a short list of rules? also, is this really needed? If we have the guard provisioning hook, shouldn't that be enough?

palazzem: Aggregated: the deny list now carries only the canonical command names — `ansible`, `ansible-playbook`, `ansible-pull`, `ansible-console`, `hanzo`, `hanzo-aur`, `shelly install` (each bare and with args), plus the pre-existing `bootstrap.sh` entries. Every tilde / absolute / `./bin` / `bash bin/...` permutation is gone; the PreToolUse hook covers those because it pattern-matches execution positions across the whole command line, so it also catches pipelines, subshells and env-assignment prefixes that a literal list can never enumerate. I re-ran the matrix after trimming: all 24 dropped permutations still exit 2, and the allow cases (`cat`, `grep`, `git`, `shellcheck`, `docker build ... -t hanzo:test`) still exit 0. I kept the thin list alongside the hook because it is declarative and self-documenting, and it still applies if hooks are disabled or the hook script breaks — hook-only would make that script a single point of failure.

PR #333, opened 2026-08-18
feat(cli): print remaining manual steps after provisioning

palazzem:
### Related Issues

- N/A

### Proposed Changes:

A provisioning run currently ends on Shelly's last AUR package, leaving no
trace of the work that automation cannot do. `bin/hanzo` now closes with a
`manual_steps` block listing the three steps that stay with the operator —
they need account secrets or firmware access no playbook can reach:

1. **ProtonVPN over WireGuard** — generate a configuration in the ProtonVPN
   account, import it with NetworkManager
   (https://protonvpn.com/support/wireguard-linux#cli)
2. **Secure Boot** — firmware into setup mode, enroll own keys with `sbctl`,
   sign the bootloader and the kernel
   (https://wiki.cachyos.org/configuration/secure_boot_setup/)
3. **Reboot** — `sudo systemctl reboot`, since kernel, firmware, and service
   changes only take effect on the next boot

The block is boxed, numbered, and colorized with the palette already used by
`bootstrap.sh` (blue rule, green step markers, cyan links and commands, dim
footer).

Gating mirrors the AUR stage: printed whenever provisioning actually ran, and
skipped under `--check`, which changed nothing and has nothing to hand back.
That keeps the path exercised by the weekly `--ci` container run.

`README.md` grows a `Manual Steps` section with the same three items and the
Quickstart list gains the matching final step.

### Testing:

- `pre-commit run --all-files` — all hooks pass (shellcheck included)
- `docker build --build-arg HANZO_ARGS="--check" -f tests/Containerfile -t hanzo:test .`
  — `ok=42 changed=24 failed=0`, image built
- Rendered `manual_steps` directly in a shell to check the layout, wrapping,
  and colors

### Extra Notes (optional):

Worth a look at the gate choice: the steps also print at the end of a `--ci`
container run, where no human reads them. That is deliberate — it keeps the
output under test — but restricting it to attended runs (`[ -z "$MODE" ]`) is
a one-line change if the noise is not wanted.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ This will:
 5. Install [ansible-core](https://docs.ansible.com/ansible-core/)
 6. Install required Galaxy collections (`community.general`)
 7. Run the full provisioning: the playbook (prompts once for your sudo password), then the AUR package set through Shelly (one PKGBUILD review per package)
+8. Print the manual steps automation cannot cover (see below)
 
 ## Usage
 
@@ -38,6 +39,14 @@ hanzo              # full provisioning run
 hanzo --check      # dry run (shows what would change)
 hanzo --ci         # unattended run, container only
 ```
+
+## Manual Steps
+
+Every provisioning run ends by printing the three steps that stay in your hands — they need account secrets or firmware access no playbook can reach:
+
+1. [ProtonVPN over WireGuard](https://protonvpn.com/support/wireguard-linux#cli) — generate a WireGuard configuration in your ProtonVPN account, then import it with NetworkManager
+2. [Secure Boot](https://wiki.cachyos.org/configuration/secure_boot_setup/) — put the firmware in setup mode, enroll your own keys with `sbctl`, then sign the bootloader and the kernel
+3. Reboot with `sudo systemctl reboot` — kernel, firmware, and service changes only take effect on the next boot
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ This will:
 5. Install [ansible-core](https://docs.ansible.com/ansible-core/)
 6. Install required Galaxy collections (`community.general`)
 7. Run the full provisioning: the playbook (prompts once for your sudo password), then the AUR package set through Shelly (one PKGBUILD review per package)
-8. Print the manual steps automation cannot cover (see below)
+8. Print manual steps automation cannot cover
 
 ## Usage
 
@@ -39,14 +39,6 @@ hanzo              # full provisioning run
 hanzo --check      # dry run (shows what would change)
 hanzo --ci         # unattended run, container only
 ```
-
-## Manual Steps
-
-Every provisioning run ends by printing the three steps that stay in your hands — they need account secrets or firmware access no playbook can reach:
-
-1. [ProtonVPN over WireGuard](https://protonvpn.com/support/wireguard-linux#cli) — generate a WireGuard configuration in your ProtonVPN account, then import it with NetworkManager
-2. [Secure Boot](https://wiki.cachyos.org/configuration/secure_boot_setup/) — put the firmware in setup mode, enroll your own keys with `sbctl`, then sign the bootloader and the kernel
-3. Reboot with `sudo systemctl reboot` — kernel, firmware, and service changes only take effect on the next boot
 
 ## Configuration
 
--- a/bin/hanzo
+++ b/bin/hanzo
@@ -9,11 +9,13 @@ ANSIBLE_CORE_VERSION="2.21.3"
 GREEN='\033[0;32m'
 BLUE='\033[0;34m'
 CYAN='\033[0;36m'
+YELLOW='\033[0;33m'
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -39,6 +40,14 @@ hanzo --check      # dry run (shows what would change)
 hanzo --ci         # unattended run, container only
 ```
 
+## Manual Steps
```
palazzem: Remove this section. It's the script that should mention what are the steps.

README.md
```diff
@@ -28,6 +28,7 @@ This will:
 5. Install [ansible-core](https://docs.ansible.com/ansible-core/)
 6. Install required Galaxy collections (`community.general`)
 7. Run the full provisioning: the playbook (prompts once for your sudo password), then the AUR package set through Shelly (one PKGBUILD review per package)
+8. Print the manual steps automation cannot cover (see below)
```
palazzem: "Print manual steps automation cannot cover"

bin/hanzo
```diff
@@ -56,4 +88,5 @@ ansible-playbook playbook.yml "${ARGS[@]}"
 # Shelly equivalent, so --check skips it.
 if [ "$MODE" != "--check" ]; then
     "$HANZO_DIR/bin/hanzo-aur" ${MODE:+"$MODE"}
+    manual_steps
```
palazzem: manual steps must be printed in all cases

README.md
palazzem: **Claude Harness**

Removed the whole `## Manual Steps` section in 19fb220. The script is now the only place the step list lives.

README.md
palazzem: **Claude Harness**

Reworded to `8. Print manual steps automation cannot cover` in 19fb220. The `(see below)` pointer is gone along with the section it referenced.

bin/hanzo
palazzem: **Claude Harness**

Moved `manual_steps` out of the `--check` guard in 19fb220 — it is now the last statement of the script and runs in every mode. Only the `hanzo-aur` call stays behind the guard.

One consequence handled with it: the banner headline read `Provisioning complete.`, which is false on a dry run that applied nothing. It is now mode-neutral — `Hanzo is done here.` — with no conditional branching in the banner and the rest of the layout unchanged.

Verified in the `--check` container build, where the banner now prints after the play recap:

```
localhost                  : ok=42   changed=24   unreachable=0    failed=0    skipped=32   rescued=0    ignored=0

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  Hanzo is done here. Three steps are still yours to run.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

bin/hanzo
```diff
@@ -57,3 +89,7 @@ ansible-playbook playbook.yml "${ARGS[@]}"
 if [ "$MODE" != "--check" ]; then
     "$HANZO_DIR/bin/hanzo-aur" ${MODE:+"$MODE"}
```
palazzem, line 92: Extend the scope: include a warning log to mention that the check skipped AUR installation because hanzo was executed with --check

palazzem: **Claude Harness**

Added in c6c42ce. The palette gains `YELLOW='\033[0;33m'` and a `warn()` helper next to `log()`/`fail()` — same `[hanzo]` prefix, stderr like `fail()`:

```bash
warn() { echo -e "${YELLOW}[hanzo]${NC} $1" >&2; }
```

The AUR gate now has an `else` branch:

```bash
if [ "$MODE" != "--check" ]; then
    "$HANZO_DIR/bin/hanzo-aur" ${MODE:+"$MODE"}
else
    warn "Skipped AUR installation: --check is a dry run."
fi
```

It renders as `[hanzo] Skipped AUR installation: --check is a dry run.` with the prefix in yellow.

Verified in the `--check` container build, where it prints right before the handoff banner:

```
localhost                  : ok=42   changed=24   unreachable=0    failed=0    skipped=32   rescued=0    ignored=0

[hanzo] Skipped AUR installation: --check is a dry run.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  Hanzo is done here. Three steps are still yours to run.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

(In the raw BuildKit log the warning lands one line lower, between the rule and the headline — stdout and stderr are separate pipes there. On a terminal both share one fd, so the order is as shown.)

`pre-commit run --all-files` passes (shellcheck included) and the `--check` build ends `ok=42 changed=24 failed=0`.

PR #334, opened 2026-08-18
feat(home): relocate cowork into ~/workspaces and seed workspace rules

palazzem:
## Summary

Moves the cowork folder from `~/cowork` to `~/workspaces/cowork` and seeds it with the workspace rules `CLAUDE.md`.

- `home_cowork_dir` default centralizes the path (`~/workspaces/cowork`); the directory task creates it with parents.
- New task deploys `roles/home/files/cowork/CLAUDE.md` into the workspace. The file is a verbatim copy of the workspace rules — no `# Managed by Hanzo` header, since Claude Code reads it as-is and the content must match the source exactly.
- The Claude Code desktop launcher (`hanzo-claude-code.desktop`) follows the folder: `Exec` working directory and `Comment` now point at `~/workspaces/cowork`.

Already-provisioned hosts keep their old `~/cowork` untouched — removing it could destroy task data, so cleanup is left to the user.

## Verification

- `pre-commit run --all-files` — all hooks pass.
- `docker build --build-arg HANZO_ARGS="--check" -f tests/Containerfile` — provisioning check passes (`failed=0`); both new tasks report changed.


roles/home/files/cowork/CLAUDE.md.template
```diff
@@ -0,0 +1,23 @@
+# Workspace Rules
```
palazzem, line 1: Do not call it `CLAUDE.md` in this repository, otherwise an agent accessing the folder would execute this as ruling. Call it `CLAUDE.md.template` or any other best practice in the industry.

palazzem: **Claude Harness**

Renamed the repo-side source to `roles/home/files/cowork/CLAUDE.md.template` via `git mv` (content byte-identical, 100% rename similarity) so an agent working in this repository will not pick it up as rules. The copy task now uses `src: cowork/CLAUDE.md.template` while `dest` stays `{{ home_cowork_dir }}/CLAUDE.md`, so the deployed file keeps its real name. Commit 5dab253.

PR #335, opened 2026-08-18
fix(home): apply PowerDevil sleep mode to LowBattery profile

palazzem:
### Related Issues

- None — bug found by inspection: sleep mode was configured for AC Power and Battery but not On Low Battery.

### Proposed Changes:

Plasma 6 PowerDevil has three power profiles: `AC`, `Battery`, and `LowBattery`. The deployed `~/.config/powerdevilrc` set `SleepMode=3` (Suspend then Hibernate) only under `[AC][SuspendAndShutdown]` and `[Battery][SuspendAndShutdown]`, so once the battery dropped into the low-battery profile the lid action fell back to plain suspend and the machine never hibernated. This adds the missing `[LowBattery][SuspendAndShutdown]` group with the same `SleepMode=3`, making the behavior consistent across all three power states.

### Testing:

- `pre-commit run --all-files` passes.
- Container provisioning check passes: `docker build --build-arg HANZO_ARGS="--check" -f tests/Containerfile -t hanzo:test .` (exit 0).

### Extra Notes (optional):

Group name verified against PowerDevil's Plasma 6 config schema (`LowBattery` is the third profile group, same `SuspendAndShutdown`/`SleepMode` keys).

### Checklist

### DIFF
--- a/roles/home/tasks/kde.yml
+++ b/roles/home/tasks/kde.yml
@@ -35,8 +35,9 @@
   become: false
 
 # KDE Plasma 6: PowerDevil writes lid-action behavior here. SleepMode=3
-# = "Suspend then Hibernate" — set under both AC and Battery so behavior
-# is consistent regardless of power source.
+# = "Suspend then Hibernate" — set under all three power profiles (AC,
+# Battery, LowBattery) so behavior is consistent regardless of power
+# source.
 - name: Configure KDE PowerDevil sleep mode
   ansible.builtin.copy:
     dest: "{{ ansible_facts.env.HOME }}/.config/powerdevilrc"
@@ -46,6 +47,9 @@
       SleepMode=3
 
       [Battery][SuspendAndShutdown]
+      SleepMode=3
+
+      [LowBattery][SuspendAndShutdown]
       SleepMode=3
     mode: "0644"
   when: host_has_kde


palazzem:
**Claude Harness**

Review-panel angle selection (deterministic, judged against the PR diff):

Core — always on:
- `correctness` — selected
- `divergence` — selected (no spec issue, no design doc: judged vs what the PR itself declares)
- `simplicity` — selected
- `settling` — selected

Conditional:
- `security` — skipped: no authn/authz, secret handling, crypto, or input parsing in the diff
- `reliability` — skipped: no process-boundary call or retry/timeout/recovery logic touched (diff is static file content in an `ansible.builtin.copy` task)
- `performance` — skipped: no loop body or caching

Repo-supplied (`.claude/review-angles/*.md`): none present.

palazzem (commented):
**Claude Harness**

Examined the full PR diff (`roles/home/tasks/kde.yml`, the `Configure KDE PowerDevil sleep mode` task) and the surrounding role for abstraction, indirection, or generality no requirement demands. The change adds one literal `[LowBattery][SuspendAndShutdown]` group to an inline `content:` block and extends the task comment — no new variables, loops, templates, or parameterization. The three repeated `SleepMode=3` lines stay as literal INI text, which is the most direct expression of the requirement; no undemanded generality was introduced.

palazzem (commented):
**Claude Harness**

Examined whether the tests cited by this PR can prove its claim — that the new `[LowBattery][SuspendAndShutdown]` group in `roles/home/tasks/kde.yml` is deployed and correct. Traced the task's `when: host_has_kde` gate through `playbook.yml` pre_tasks and the container test environment (`tests/Containerfile`, `cachyos/cachyos:latest`) to determine whether the changed task is ever evaluated by `pre-commit` or the container build. No spec issue or design doc exists; reviewed against the PR body's Testing section and checklist.

- `testing-quality/F1` — Both cited tests skip the changed task: the container has no plasmashell, so `host_has_kde` is false and a passing build proves nothing about the new group → roles/home/tasks/kde.yml:52

roles/home/tasks/kde.yml
```diff
@@ -47,6 +48,9 @@
 
       [Battery][SuspendAndShutdown]
       SleepMode=3
+
+      [LowBattery][SuspendAndShutdown]
```
palazzem, line 52: **Claude Harness**

**`testing-quality/F1` — The tests cited as evidence never evaluate this change: the container test skips the task, so a passing build cannot fail on any content of this block**

- **Evidence:** This task is gated `when: host_has_kde` (roles/home/tasks/kde.yml:55), set in `playbook.yml:34-47` from `stat /usr/bin/plasmashell` during `pre_tasks` — before any role runs. `tests/Containerfile` installs only `python git curl sudo base-devel`, and `docker run --rm cachyos/cachyos:latest sh -c 'test -e /usr/bin/plasmashell && echo PLASMASHELL_PRESENT || echo PLASMASHELL_ABSENT'` prints `PLASMASHELL_ABSENT`; `grep -rn "plasmashell\|plasma-desktop" roles/ tests/ bin/` matches nothing that provides it. So under both `--check` and `--ci` the task is skipped, and `pre-commit` only lints YAML — neither cited test parses the INI content. The build would pass identically with `[LowBatery]`, a wrong key, or this change reverted.
- **Impact:** The PR's Testing section ("Container provisioning check passes") and the checked box "Tests are defining the correct and expected behavior" read as verification of this fix, but no test in the run can fail if the fix is wrong — the same gap that let the original AC/Battery-only config ship unnoticed. The `[LowBattery]` group name and `SleepMode=3` value rest entirely on the PR-body assertion, with no assertion anywhere that can catch a regression in `host_has_kde`-gated content.

palazzem (commented):
**Claude Harness**

Examined the single changed task in `roles/home/tasks/kde.yml` ("Configure KDE PowerDevil sleep mode"): the added `[LowBattery][SuspendAndShutdown]` group with `SleepMode=3` and the updated comment. Verified every load-bearing claim against upstream PowerDevil source: profile IDs are exactly `AC`, `Battery`, `LowBattery` (`daemon/powerdevilcore.cpp`, `profileId = QStringLiteral("LowBattery")`); `SleepMode` is a UInt entry in the `[$(ProfileId)][SuspendAndShutdown]` group of `powerdevilrc` (`PowerDevilProfileSettings.kcfg`); enum value 3 is `SleepMode::SuspendThenHibernate` (`daemon/powerdevilenums.h`); and the pre-PR fallback claim holds — the kcfg default is `SleepMode::SuspendToRam`, so the low-battery profile previously fell back to plain suspend. The deployed content is well-formed KConfig INI matching real Plasma 6 `powerdevilrc` files, and `pre-commit run --all-files` (yamllint, ansible-lint) passes on the PR branch in an isolated worktree. The diff does what it claims; no logic, boundary, or error-path defects found.

roles/home/tasks/kde.yml
palazzem, line 52: **Claude Harness**

**`settling/F1` — The fix extends the whole-file `copy` overwrite of `powerdevilrc` — a convention that deletes every user-made power setting on each run — when the same task file already demonstrates the correct tool for this exact job.**

- **Evidence:** `ansible.builtin.copy` with `content:` replaces the destination file wholesale on every apply, and Hanzo re-applies weekly (CLAUDE.md: `--ci` is "what the weekly cron runs"). Plasma 6 writes user power-management changes into this very file — KDE's PowerDevil 6.0 announcement: "Power management settings are now only written if you change away from the default values… Extra config files for the various power states and activities have been migrated to `~/.config/powerdevilrc`" (https://blogs.kde.org/2024/04/23/powerdevil-in-plasma-6.0-and-beyond/). So any System Settings change (dim/screen-off timeouts, battery thresholds, power-button action) lands in `powerdevilrc` and is erased by the next provisioning run. The better approach is evident in this same file: lines 26–35 use `community.general.ini_file` with `section: "services][{{ … }}"` and a comment explaining the `]`+`[` nested-group syntax — exactly what PowerDevil's `[AC][SuspendAndShutdown]` groups need. A single `ini_file` task looped over `[AC, Battery, LowBattery]` sets `SleepMode=3` while leaving every other key intact, and dissolves the block this diff duplicates for the third time. Precedent for a Hanzo-managed keyed file without the full-ownership header already exists in the `kglobalshortcutsrc` task above. This diff instead grows the overwrite payload and rewrites the comment to justify the pattern ("set under all three power profiles") — settling the flawed convention deeper.
- **Impact:** Every weekly provisioning run silently reverts all user PowerDevil settings except `SleepMode` to defaults, and each future profile or key grows a copy-pasted inline config instead of one looped keyed task.

palazzem (commented):
**Claude Harness**

**Angle `settling` — 1 finding**

Examined whether this diff extends a convention the codebase should be moving away from instead of into: the whole-file `ansible.builtin.copy` overwrite of `~/.config/powerdevilrc` in `roles/home/tasks/kde.yml`, which this change grows by a third duplicated `SleepMode=3` block, versus the keyed `community.general.ini_file` approach already used — nested-group trick included — in the task directly above it.

- `settling/F1` — Extending the whole-file overwrite of `powerdevilrc` entrenches a pattern that wipes user power settings on every run; the evident alternative (`ini_file` with the nested-group trick) is already used 15 lines up in the same file → roles/home/tasks/kde.yml:52

palazzem (commented):
**Claude Harness**

Examined the sole changed file `roles/home/tasks/kde.yml` against the PR's own declarations (no spec issue, no design doc exist for this PR). Verified: (1) the diff adds exactly the declared `[LowBattery][SuspendAndShutdown]` group with `SleepMode=3` and updates the task comment to match, with no undeclared changes; (2) the declared config schema holds against current PowerDevil source — profile groups `AC`/`Battery`/`LowBattery` in `daemon/powerdevilcore.cpp`, `SleepMode` enum with `SuspendThenHibernate = 3` in `daemon/powerdevilenums.h`, and the declared plain-suspend fallback confirmed by the kcfg default `SuspendToRam` for `SleepMode` in `PowerDevilProfileSettings.kcfg` (group `SuspendAndShutdown`, parent `$(ProfileId)`); (3) both declared test results reproduced in an isolated worktree of the PR branch — `pre-commit run --all-files` all hooks passed, and `docker build --build-arg HANZO_ARGS="--check" -f tests/Containerfile` exited 0 with `PLAY RECAP failed=0`. No departure, omission, or invention relative to what the PR declares.

PR #337, opened 2026-08-20
feat(kde): provision KDE Plasma configuration as a role

palazzem:
Integrates the KDE Plasma 6 setup captured in `var/kde-provisioning/` into hanzo provisioning, and lands the accompanying investigation notes as an official repository skill.

## What changed

**New `kde` role** (tag `kde`, gated on `host_has_kde`). Every setting applies **live against the running Plasma session** during a `hanzo` run — nothing is deferred to the next login. Settings take one of three shapes:

| Shape | Used for |
|---|---|
| `ini_file` key + reload handler | screen edges, touchscreen gestures, Darkly decorations |
| Command guarded by a `kreadconfig6` read | Darkly color scheme, Plasma style, widget style |
| Script in `files/` | touchpad, virtual desktops, Meta+N shortcuts, display, virtual keyboard, panel layout |

Scripts print `CHANGED` only when they actually wrote something, so re-runs are quiet. That matters most for the panel layout, which removes every existing panel before rebuilding — deploying `panel-layout.js` is the change signal, so an unchanged run no longer tears down the panel or restarts plasmashell.

The KDE tasks that lived in the home role (Claude Code launcher, F23 Copilot key, PowerDevil sleep mode) move here, so the domain lives in one place per rule 4.

**`bin/hanzo` runs `hanzo-aur` before the playbook.** AUR is the *attended* half of a run — one PKGBUILD review per package — so the reorder puts every prompt up front and lets the playbook then run unattended to the end. It also resolves the ordering problem that blocked this work: Darkly is AUR-only (not in any CachyOS binary repo), so the role that themes with it needs the package already on disk. `hanzo-aur` keeps aborting the run on failure — all AUR packages must install. Shelly moves from `system_pacman` to the CLI, which needs it before a role could install it, and installs with `-Sy` because the playbook's database refresh now runs afterward.

Darkly theming stays gated on the package being present — declining a PKGBUILD review is a legitimate outcome and must not fail the run.

**`.claude/skills/kde-config/`** — the eight lesson files carry over intact (they document KDE primitives, which the move to Ansible does not change), with cross-references rewired from the standalone scripts to the tasks and handlers that replaced them. `SKILL.md` is rewritten around the role.

## Verification

- `pre-commit run --all-files` — passes (ansible-lint, shellcheck, codespell).
- `HANZO_ARGS="--check"` container build — passes (41 ok, 23 changed, 0 failed).
- `HANZO_ARGS="--ci"` container build — the reordered flow was confirmed running Shelly first (shelly installed, 1Password PGP-verified and built, PKGBUILD prompts auto-accepted) before any Ansible output. The full run is slow to complete under x86 emulation locally; the weekly workflow exercises it natively.

**CI cannot verify the role itself.** The test container has no Plasma, so `host_has_kde` is false and every KDE task is skipped — the builds above prove syntax, lint and the reordered CLI, nothing about the settings. Real verification is a run on the machine:

- [ ] `hanzo` applies each setting live: touchpad, 5 virtual desktops + Meta+1..5, display scale/VRR/rotate, virtual keyboard, no Overview hot corner, bottom-edge swipe → Overview, top-edge swipe → launcher, panel layout, Darkly across all four components.
- [ ] A second `hanzo` run reports **no changes** for the kde role — and in particular does not rebuild the panel or restart plasmashell.
- [ ] `ansible.builtin.script` + `environment:` passes the tunables through as expected (unexercised in CI).
- [ ] The reordered run feels right in practice: PKGBUILD reviews up front, then unattended to the end.

Once verified on the machine, `var/kde-provisioning/` can be retired — it is its own git repository, so its history survives outside hanzo.


.claude/skills/kde-config/SKILL.md
```diff
@@ -0,0 +1,229 @@
+---
+name: kde-config
+description: Add or change a KDE Plasma 6 setting in Hanzo's kde role (touchpad, virtual desktops, global shortcuts, display, Darkly theme, panel layout, virtual keyboard, screen edges, touchscreen gestures, PowerDevil). Use when a new Plasma setting must be provisioned, or an existing one fixed or extended, without re-deriving the underlying D-Bus/config knowledge from scratch.
```
palazzem: Remove `(touchpad, virtual desktops, global shortcuts, display, Darkly theme, panel layout, virtual keyboard, screen edges, touchscreen gestures, PowerDevil)` as it's enough to say KDE role.

.claude/skills/kde-config/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+
+This is the escape hatch (`CLAUDE.md` rule 1) — justified only when no
+module and no simple guarded command can do it. When a tool is a
+declarative apply with no change signal at all (`kscreen-doctor`), mark
+the task `changed_when: false` and say why in a comment; do not invent a
+comparison layer.
+
+## Repository rules that apply here
+
+- **Never run `ansible-playbook`, `hanzo`, or `bootstrap.sh` on the host.**
```
palazzem: Remove this rule as it's enforced via pre-hook and CLAUDE.md

.claude/skills/kde-config/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+
+```bash
+# list a directory
+curl -s "https://invent.kde.org/api/v4/projects/<namespace>%2F<repo>/repository/tree?path=<path>&ref=master&per_page=100"
+
+# read a file
+curl -s "https://invent.kde.org/<namespace>/<repo>/-/raw/master/<path>"
+```
+
+Repos used so far (namespace/repo): `plasma/plasma-desktop` (touchpad/mouse
```
palazzem: Transform this in a bullet list.

bin/hanzo
```diff
@@ -64,6 +64,27 @@ if [ -z "$MODE" ]; then
     git reset --hard origin/main
 fi
 
+# AUR provisioning runs first: it is the attended half of a run (one
```
palazzem: You are too verbose. Remove the prose and go straight to the point of Why we're doing it. Here you must just describe the execution order and the dependency.

bin/hanzo
```diff
@@ -64,6 +64,27 @@ if [ -z "$MODE" ]; then
...
+# playbook then runs unattended to the end. It also puts AUR packages on
+# disk before the roles that configure them — the kde role themes with
+# Darkly, which only exists here. A dry run has no Shelly equivalent, so
+# --check skips it.
+#
+# Shelly comes from pacman, not from a role: hanzo-aur needs it before
+# the playbook could install it. Same ownership rule as bootstrap/uv and
+# hanzo/ansible-core — whoever needs a tool first installs it.
+if [ "$MODE" != "--check" ]; then
+    # -Sy, not -S: the playbook's own database refresh now runs after this
```
palazzem: Unneeded if you write right the comment above.

roles/kde/defaults/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+# Touchpad properties, applied over the KWin InputDevice D-Bus API.
+# busctl booleans: quoted lowercase true|false.
+kde_touchpad_natural_scroll: "true"
+kde_touchpad_tap_to_click: "false"
+
+kde_virtual_desktops: 5
+kde_virtual_desktop_rows: 1
+
+# Applied to every connected+enabled output via kscreen-doctor.
```
palazzem: That's not the way to write comments. You can write one comment that explains what the setting is only if you split it by blocks (e.g. a comment -> a group of settings which the comment belongs). So if it's a Virtual desktop block, you may write a comment about Virtual desktops configs. No inline comments where you write all other possible options (it's not a documentation file this one).

roles/kde/files/panel-layout.js
```diff
@@ -0,0 +1,58 @@
...
+clock.writeConfig("autoFontAndSize", false);
+clock.writeConfig("boldText", true);
+clock.writeConfig("firstDayOfWeek", 1);
+clock.writeConfig("fontFamily", "DejaVu Sans Mono");
+clock.writeConfig("fontSize", 8);
+clock.writeConfig("fontStyleName", "Bold");
+clock.writeConfig("fontWeight", 700);
+clock.writeConfig("showWeekNumbers", true);
+
+// Not added: org.kde.plasma.pager (redundant with the Meta+1..N desktop
```
palazzem: Unneeded comment. remove such kind of comments as we don't care about "what we didn't add".

roles/kde/templates/hanzo-claude-code.desktop.j2
```diff
@@ -3,7 +3,7 @@
 Type=Application
 Name=Claude Code
 Comment=Open Claude Code in ~/workspaces/cowork
-Exec=ghostty --working-directory={{ home_cowork_dir }} -e fish -c claude
+Exec=ghostty --working-directory={{ kde_cowork_dir }} -e fish -c claude
```
palazzem: this has nothing to do with KDE. this variable must be called "cowork_dir". Apply everywhere and detach the two.

roles/kde/vars/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+# KDE resolves the global shortcut to the launcher by this filename, so the
+# deployed .desktop file and the kglobalshortcutsrc group must match exactly.
+# Renaming it in only one place silently kills the Copilot key.
+kde_claude_launcher_file: hanzo-claude-code.desktop
```
palazzem: this has nothing to do with KDE. why you made such changes?

CLAUDE.md
```diff
@@ -10,7 +10,7 @@ CachyOS system provisioner powered by Ansible.
...
 6. Registered variables inside a role must use the role name as prefix (ansible-lint `var-naming[no-role-prefix]` rule). E.g., inside `roles/tools` use `tools_uv_tool_list`, not `uv_tool_list`.
-7. AUR packages are never managed by Ansible. The package set lives in `bin/hanzo-aur`, which installs it via `shelly install aur` — Shelly shows each PKGBUILD diff for human review before building, and packages with signing upstreams (1Password) are PGP-verified at a pinned AUR commit first. `bin/hanzo` runs it after the playbook — `--check` skips it, `--ci` (container only) auto-accepts Shelly's prompts.
+7. AUR packages are never managed by Ansible. The package set lives in `bin/hanzo-aur`, which installs it via `shelly install aur` — Shelly shows each PKGBUILD diff for human review before building, and packages with signing upstreams (1Password) are PGP-verified at a pinned AUR commit first. `bin/hanzo` runs it *before* the playbook, so the attended PKGBUILD reviews happen up front and the playbook then runs unattended; roles may therefore configure AUR packages (the `kde` role themes with Darkly). `--check` skips it, `--ci` (container only) auto-accepts Shelly's prompts.
```
palazzem: You are putting two rules in the same line. Split it.

README.md
```diff
@@ -27,7 +27,7 @@ This will:
 4. Link `hanzo` and `hanzo-aur` into `~/.local/bin`
 5. Install [ansible-core](https://docs.ansible.com/ansible-core/)
 6. Install required Galaxy collections (`community.general`)
-7. Run the full provisioning: the playbook (prompts once for your sudo password), then the AUR package set through Shelly (one PKGBUILD review per package)
+7. Run the full provisioning: the AUR package set through Shelly (one PKGBUILD review per package), then the playbook (prompts once for your sudo password)
```
palazzem: You are putting two steps in the same line.

README.md
```diff
@@ -55,7 +55,7 @@ The file is loaded into its own namespace and read through an allowlist: exactly
 
 ## Architecture
 
-Hanzo provisions the local machine with `ansible-playbook playbook.yml`, then hands the AUR package set to Shelly — Ansible never manages AUR. All operations are idempotent — running `hanzo` multiple times is safe and will only apply changes that are needed.
+Hanzo hands the AUR package set to Shelly first — Ansible never manages AUR — and then provisions the local machine with `ansible-playbook playbook.yml`. AUR goes first because it is the attended half of a run: every PKGBUILD review happens up front, the playbook then runs unattended to the end, and roles that configure AUR packages (the `kde` role themes with Darkly) find them already installed. All operations are idempotent — running `hanzo` multiple times is safe and will only apply changes that are needed.
```
palazzem: This sentence is bad. Rewrite it entirely as this should describe briefly the architecture. Keep OSS standards.

.claude/skills/kde-config/SKILL.md
palazzem: Done — description now stops at the role.

.claude/skills/kde-config/SKILL.md
palazzem: Removed.

.claude/skills/kde-config/SKILL.md
palazzem: Done — the repo list is a bullet list now, with the GitHub note as its own paragraph.

bin/hanzo
palazzem: Rewritten to state the order and the dependency only: shelly is installed here because hanzo-aur needs it before the playbook could provide it (with -Sy, since the database refresh now runs after this step), AUR precedes the playbook, --check skips it.

bin/hanzo
palazzem: Right — folded into the comment above and removed.

roles/kde/defaults/main.yml
palazzem: Reworked into blocks: touchpad, virtual desktops, displays, virtual keyboard, screen edges and gestures, Darkly, panel — one comment per block, no per-line enumeration of accepted values. The one value note kept is that the border settings take ElectricBorder integers, because a symbolic name there silently parses as 0.

roles/kde/files/panel-layout.js
palazzem: Removed, along with the same note in the file header.

roles/kde/templates/hanzo-claude-code.desktop.j2
palazzem: Agreed, and it went further than a rename. The .desktop launcher is desktop-agnostic, so it moved back to the home role with the template; the kde role keeps only the Copilot-key binding, which is genuinely Plasma (it writes kglobalshortcutsrc). The variable is now `cowork_dir` at play scope in playbook.yml, so home and kde share one name instead of each carrying a role-prefixed copy — `home_cowork_dir` is gone too.

roles/kde/vars/main.yml
palazzem: You're right — that was a bad call on my part. I moved the launcher into the kde role while consolidating the KDE domain and renamed the variable to match the role prefix, but the launcher itself has nothing to do with KDE. It is back in the home role, the shared name is `claude_launcher_file` at play scope, and this `vars/main.yml` is deleted — the kde role no longer has one.

CLAUDE.md
palazzem: Split: rule 7 keeps the AUR-is-never-Ansible contract and Shelly's review/verification, and a new rule 8 states the ordering and what it enables.

README.md
palazzem: Split into step 7 (AUR package set through Shelly) and step 8 (run the playbook); the manual-steps item is now 9.

README.md
palazzem: Rewritten as a two-stage description: Shelly installs the AUR package set first, pausing on each PKGBUILD for review, and the playbook then provisions the machine unattended. It then states that Ansible never manages AUR packages, that the ordering puts every prompt up front and lets roles configure what Shelly installed, and that runs are idempotent. See the Architecture section in the updated README.

CLAUDE.md
palazzem: Rework from scratch Rules section. These are rules, you are adding implementation details like check skips AUR package and so on. Task a Fable subagent to read this file and rework this section so that generic rules are truly extrapolated.

README.md
```diff
@@ -55,7 +56,7 @@ The file is loaded into its own namespace and read through an allowlist: exactly
 
 ## Architecture
 
-Hanzo provisions the local machine with `ansible-playbook playbook.yml`, then hands the AUR package set to Shelly — Ansible never manages AUR. All operations are idempotent — running `hanzo` multiple times is safe and will only apply changes that are needed.
+A run has two stages. Shelly installs the AUR package set first, pausing on each PKGBUILD for review; `ansible-playbook playbook.yml` then provisions the machine unattended. Ansible never manages AUR packages, and the ordering means every prompt comes up front and roles can configure what Shelly installed. All operations are idempotent — running `hanzo` multiple times is safe and applies only what changed.
```
palazzem: This is bad as it doesn't describe the architecture, it tells you implementation details. Architecture IS the architecture, so no implementation details. You must explain how this project is organized and what is the execution order. Period. If there are caveats or other things, that's an implementation detail that is clearly described in the code.

CLAUDE.md
palazzem: Reworked from scratch by a Fable subagent, which audited all eight rules rather than just the two I had touched. Result is seven, each a constraint on authoring rather than a description of current behavior.

Removed as implementation detail: mode handling (which mode skips or auto-accepts AUR), Shelly's PKGBUILD-diff and PGP-pinning mechanics, the ordering rationale narration, the kde/Darkly example, and rule 1's fnm aside.

Old rules 7 and 8 collapsed into one, since they shared a single principle: AUR is never Ansible's, every package gets human review, and it installs before the playbook — so roles may configure AUR packages but never install them.

Rule 6 also widened while it was open: it now covers every role-scoped variable rather than only registered ones (which is how the repo already names its defaults), and records that variables shared by more than one role belong at play scope. Rules 2-5 were already generic and are untouched.

README.md
palazzem: Agreed — it was describing behavior, not structure. Rewritten to the execution order and the layout, and nothing else: two stages (AUR set through Shelly, then the playbook applying every role in dependency order), then the directory list with bin/ first since that is where a run starts.

Dropped: the idempotence claim (already stated in the intro), the Ansible-never-manages-AUR boundary (that is a rule, and lives in CLAUDE.md), and the hardware role's dispatch-and-skip condition, which is documented where it is implemented.

PR #339, opened 2026-08-25
feat(kde): bind Meta+Shift+4/5 to Spectacle captures via kglobalaccel

palazzem:
### Related Issues

No related issue.

### Proposed Changes:

Adds two global shortcuts through `kglobalaccel`: Meta+Shift+4 opens Spectacle's rectangular region capture (`RectangularRegionScreenShot`), Meta+Shift+5 captures the current monitor (`CurrentMonitorScreenShot`). Both are added as alternatives, so Spectacle's existing Print-based defaults and KWin's Ctrl+F1 defaults keep working. A pre-flight scan of all 20 kglobalaccel components (`allShortcutInfos`) confirmed neither combination was bound before this change.

In kglobalaccel terms the keys are `Meta+$` and `Meta+%`, not `Meta+Shift+4` and `Meta+Shift+5`. `Xkb::modifiersRelevantForGlobalShortcuts` (`plasma/kwin`, `src/xkb.cpp`) subtracts the modifiers xkb consumed to produce the keysym, and Shift is consumed turning `4` into `$` on a US layout; the exception that keeps Shift (BUG 370341) applies only to letters. The key itself is `QXkbCommon::keysymToQtKey` of the shifted keysym, so KWin asks for `Meta+$` (`0x10000024`). KDE's own shortcut recorder stores Shift+digit the same way (`isShiftAsModifierAllowed` in `frameworks/kguiaddons`, `kkeysequencerecorder.cpp`). A literal `Meta+Shift+4` binding is accepted and persisted by the daemon but never fires, which the first version of this branch confirmed the hard way.

The bindings are applied over D-Bus rather than by writing `~/.config/kglobalshortcutsrc` directly. `plasma/kglobalacceld`'s `GlobalShortcutsRegistry` has no `KConfigWatcher`, so a hand-edited config file is not picked up until the next login, and `KServiceActionComponent::writeSettings` deletes and rewrites each loaded component's group from memory on every scheduled write, so a hand-written entry can be silently clobbered. `org.kde.KGlobalAccel.setForeignShortcutKeys` replaces the action's key set live and the daemon persists the result on its own — observed `[services][org.kde.spectacle.desktop]` gaining `CurrentMonitorScreenShot=Meta+%` and `RectangularRegionScreenShot=Meta+$\tMeta+Shift+Print` within a minute of the call. On Plasma 6 Wayland, `org.kde.kglobalaccel` is owned by `kwin_wayland` itself, which is why the fixed 4-int key marshalling from the previous script is kept: a shorter array desyncs the D-Bus stream and crashes the compositor.

`roles/kde/files/bind-desktop-shortcuts.py` only knew how to bind Meta+N to KWin's desktop-switch actions. Rather than add a near-duplicate script for the Spectacle bindings, it is generalized into `roles/kde/files/bind-shortcuts.py`, driven by a `BINDINGS` JSON list of `{component, action, key}` objects. Keys are symbolic (`Meta+$`, `Meta+1`) and parsed from a small explicit table (Meta/Ctrl/Shift/Alt plus one printable ASCII character, whose `Qt::Key` value is its code point, checked against `/usr/include/qt6/QtCore/qnamespace.h`); an unknown component, action, modifier or key now fails the run instead of printing a warning, and `Shift` combined with anything but a letter or Space is rejected with a message pointing at the shifted character, mirroring KDE's recorder rule. Bindings stay additive: each key is appended to the action's existing alternatives rather than replacing them. The Meta+N desktop bindings are built by a `set_fact` loop over `kde_virtual_desktops`; the two Spectacle entries live in the new `kde_global_shortcuts` variable in `roles/kde/defaults/main.yml`, so further global shortcuts can be added there without touching the script.

The `kde-config` skill's shortcuts lesson is updated with the live-path-vs-config-file findings and the service-component (launcher) D-Bus shape, so the next change in this area starts from verified facts instead of re-deriving them.

### Testing:

- `pre-commit run --all-files` — all hooks passed (ansible-lint, shellcheck, codespell, actionlint, gitleaks, generic hooks).
- `docker build --build-arg HANZO_ARGS="--check" -f tests/Containerfile -t hanzo:test .` — exit 0, `PLAY RECAP: ok=43 changed=25 unreachable=0 failed=0 skipped=55`.
- CI cannot exercise the `kde` role: the container has no Plasma, so `host_has_kde` is false and the role is skipped. Verification below was done live on the author's Plasma 6.7.4 session (`kwin_wayland` PID 1787 unchanged throughout) and needs a `hanzo` run on the actual machine after merge, with a second run expected to report the "Bind global shortcuts" task unchanged.
  - The add path was exercised by the first version of the branch: running `roles/kde/files/bind-shortcuts.py` with the full binding set (Meta+1..5 plus the two Spectacle entries) printed exactly one `CHANGED: org.kde.spectacle.desktop/RectangularRegionScreenShot += ...` line for the missing key, with Meta+1..5 already present. Those literal `Meta+Shift+4/5` alternatives never fired; they were replaced live with `Meta+$` and `Meta+%` through the same `setForeignShortcutKeys` call, after which both physical chords (Win+Shift+4, Win+Shift+5) trigger Spectacle. Running the script with the final binding set then printed nothing and exited 0.
  - `allShortcutInfos` afterwards: `RectangularRegionScreenShot` active keys `[0x13000009, 0x10000024]`, `CurrentMonitorScreenShot` active keys `[0x10000025]`.
  - Failure paths, all exit 1 without any D-Bus write: unknown component → `ERROR: getComponent failed: Call failed: The component 'does.not.exist' doesn't exist.`; unknown action → `ERROR: action 'Nope' not found in component 'org.kde.spectacle.desktop'`; unknown modifier → `ERROR: unknown modifier 'Super' in 'Super+5'`; unsupported key → `ERROR: unsupported key 'Print' in 'Meta+Print'`; non-ASCII key → `ERROR: unsupported key 'é' in 'Meta+é'`; empty key (`Meta++`) → `ERROR: unsupported key '' in 'Meta++'`; `Meta+Shift+4` and `Meta+Shift+$` → `ERROR: Shift+4 in 'Meta+Shift+4' can never match; bind the character Shift+4 produces instead`.
  - Templating check inside the built image (ansible-core 2.21.3): a playbook running the real "Build Meta+N virtual desktop shortcut bindings" task with `kde_virtual_desktops: 3` rendered a list of three mappings, and a `command` task with the same `environment: BINDINGS: "{{ (kde_desktop_shortcuts + kde_global_shortcuts) | to_json }}"` received a JSON string that `json.loads` parsed into 4 entries (three desktop bindings plus the test playbook's sample Spectacle entry).

### Extra Notes (optional):

Two warnings appear in the container build log that are not from this change and cannot be fixed here: docker's `FromPlatformFlagConstDisallowed` on `tests/Containerfile` line 1, and pacman's `warning: could not isolate the network for "/usr/bin/ldconfig"` / `error: command failed to execute correctly` from package hooks that run inside `docker build` while installing the base toolchain and `ansible-core` — steps untouched by this PR. `pre-commit run --all-files` itself reports zero warnings.

### Checklist


roles/kde/defaults/main.yml
```diff
@@ -8,6 +8,21 @@ kde_touchpad_click_method: clickfinger
 kde_virtual_desktops: 5
 kde_virtual_desktop_rows: 1
 
+# Global shortcuts, each added as an alternative to the action's existing
```
palazzem: Do not add such comments, respect what the rest of the file does

.claude/skills/kde-config/lessons/shortcuts.md
```diff
@@ -1,6 +1,69 @@
...
+`CurrentMonitorScreenShot=Meta+Shift+5` within a minute.
+
+The one legitimate config-file write is a `[services][<file>.desktop]`
+group for a launcher kglobalaccel has never seen (the Copilot key):
+`loadSettings` creates the component from that group at daemon start, and
+`setForeignShortcutKeys` cannot target a component that does not exist.
+
+## Shift+digit and Shift+symbol never match: bind the shifted character
+
+Confirmed live: `Meta+Shift+4` and `Meta+Shift+5` set through
```
palazzem, line 34: We ended up that Meta wasn't working. Didn't we move to Win key? Or was the fact it wasn't working another issue?

roles/kde/defaults/main.yml
palazzem: Trimmed to the one-line section comment the rest of the file uses (82de8f3). The explanation of why the keys are spelled `Meta+$` / `Meta+%` lives in the kde-config shortcuts lesson instead.

.claude/skills/kde-config/lessons/shortcuts.md
palazzem: It was another issue; the modifier never changed. `Meta` is KDE's name for the Windows/Super key: KWin resolves `Qt::MetaModifier` from xkb's `Mod4` (`plasma/kwin` `src/xkb.cpp:673`, `m_metaModifier = xkb_keymap_mod_get_index(m_keymap, XKB_MOD_NAME_LOGO)`, and `XKB_MOD_NAME_LOGO` is `"Mod4"` in `xkbcommon-names.h`), and the working Meta+1..5 desktop shortcuts use that same `0x10000000` constant.

What failed was the key half of the combination. With Shift held, xkb turns `4` into `$` and marks Shift as consumed; `Xkb::modifiersRelevantForGlobalShortcuts` strips consumed modifiers (letters excepted, BUG 370341), so KWin asks kglobalaccel for `Meta+$`, never `Meta+Shift+4`. The fix in 007cc19 binds `Meta+$` and `Meta+%` with the same Meta modifier. The lesson now states this up front and the constants table explains the Meta/Windows mapping (1169172).

PR #360, opened 2026-09-16
feat(kde): open ghostty windows fullscreen instead of at a fixed size

palazzem:
### Related Issues

No related issue. Requested by the maintainer: ghostty should start fullscreen instead of at a fixed number of pixels.

### Proposed Changes:

The `hanzo-ghostty` KWin window rule (`kde_window_rules` in `roles/kde/defaults/main.yml`) applied a 1600x1001 size to every new ghostty window. It now applies fullscreen initially: `fullscreen=true` with `fullscreenrule=3` ("Apply initially"), the same thing System Settings > Window Rules writes when the Fullscreen property is enabled with that rule type.

Both keys come from source, not guesswork: `fullscreen` is a `Bool` and `fullscreenrule` a `Rules::SetRule` int in plasma/kwin `src/rulesettings.kcfg`. `XdgToplevelWindow::initialize()` in `src/xdgshellwindow.cpp` applies initial rules in a fixed order (size, then maximize, then fullscreen), so a fullscreen rule wins at mapping time.

The size keys are simply gone from the rule. `community.general.ini_file` only writes the keys it is given, so a machine provisioned with the old rule keeps `size`/`sizerule` in `kwinrulesrc`; that is by design. Provisioning targets a fresh, clean machine and is never backward compatible, now stated as rule 8 in `CLAUDE.md`, so there is no cleanup task. On an existing machine, remove the leftover Size property from the `hanzo-ghostty` rule in System Settings > Window Rules if you want it gone.

The `kde-config` skill's `lessons/window-rules.md` gained the fullscreen and maximize keys, the initial-state order, and a note that a dropped property stays on an already-provisioned machine.

### Testing:

- `pre-commit run --all-files`: every hook passed, including ansible-lint.
- The container `--check` build was **not** run locally: the Docker daemon is not running on the authoring host. CI runs it. The `kde` role is skipped in the container regardless (no Plasma, `host_has_kde` is false), so the container build proves syntax for this change, not behaviour.
- Live verification is owed on the Plasma machine: run `hanzo`, expect the window-rule task to report `fullscreen` and `fullscreenrule` changed and the `Reconfigure KWin` handler to fire; open a **new** ghostty window (an `Apply` rule only affects windows mapped after the reload) and confirm it is fullscreen; run `hanzo` again and confirm no change.

### Extra Notes (optional):

Fullscreen and maximized are distinct KWin properties: fullscreen hides the panel and the title bar, maximized keeps both. If maximized is what is wanted, the rule takes `maximizevert`/`maximizehoriz` with their `*rule` companions instead.

If the hand-made UUID rule from #340 is still present, it keeps applying its size while this rule applies fullscreen; the properties do not overlap, so both take effect.

### Checklist

- [ ] Related issue and proposed changes are filled: no related issue exists; proposed changes above stand in.
- [ ] Tests are defining the correct and expected behavior: no test harness exists for role behaviour beyond the container `--check` build and the manual Plasma verification described above.
- [ ] Container test passes: `docker build -f tests/Containerfile -t hanzo:test .`: not run locally (Docker daemon not running); CI runs it, and the `kde` role is skipped in the container regardless.


roles/kde/defaults/main.yml
```diff
@@ -47,16 +47,19 @@ kde_panel_layout_file: "{{ ansible_facts.env.HOME }}/.local/share/hanzo/panel-la
...
     Description: Window settings for com.mitchellh.ghostty
     wmclass: ghostty com.mitchellh.ghostty
     wmclasscomplete: "true"
     wmclassmatch: 1
     types: 1
-    size: 1600,1001
-    sizerule: 3
+    fullscreen: "true"
+    fullscreenrule: 3
+    size: ~
```
palazzem: Do you need to declare these with `~`? If it's for backward compatibility, add a CLAUDE.md rule to never be backward compatible as I want always a fresh and clean provisioning.

palazzem: Yes, they existed only for backward compatibility: a `~` entry made the task delete `size`/`sizerule` from a `kwinrulesrc` that an earlier run had written. Dropped in 7edcfdd together with the `state: absent` line, so the rule now lists only the keys it sets. `CLAUDE.md` gained rule 8: never be backward compatible, provisioning targets a fresh, clean machine. The lesson file now says a dropped property stays on an already-provisioned machine by design, and the PR body was updated to match.

— Claude

CLAUDE.md
```diff
@@ -11,6 +11,7 @@ CachyOS system provisioner powered by Ansible.
 5. Don't install packages already in the CachyOS base image. Verify against `cachyos/cachyos:latest` (`pacman -Qe`) before adding to a role's package list.
 6. Variables defined inside a role use the role name as prefix (ansible-lint `var-naming[no-role-prefix]` rule). E.g., inside `roles/tools` use `tools_uv_tool_list`, not `uv_tool_list`. Variables shared by multiple roles are defined at play scope in `playbook.yml`, never in one role.
 7. AUR packages are never managed by Ansible: the set lives in `bin/hanzo-aur` and installs through Shelly, with human review of every package. AUR installation precedes the playbook, so roles may configure AUR packages but never install them.
+8. Never be backward compatible. Provisioning targets a fresh, clean machine, so a change replaces what it supersedes outright: no migration or cleanup tasks for state an earlier version wrote, and no compatibility shims for existing installs.
```
palazzem, line 14: Include another rule: when a PR with a change is merged, the agent should ask confirmation to manually apply the change. It means the agent doesn't use `hanzo` (it can't) but instead alters the system manually with the user confirmation. It must describe step by step the plan of the change before executing. Transform this logic in a rule as point (9)

palazzem: Added as rule 9: when a PR that changes provisioned state is merged, the agent asks whether to apply it to the current machine by hand, reproduces the end state directly since `hanzo` never runs on the host, and describes the plan step by step before executing only what the user confirmed.

— Claude

PR #362, opened 2026-09-17
docs(kde-config): note that a bound launcher's Exec applies at next login

palazzem:
## Summary

Documents in the kde-config skill why a changed Copilot-key launcher keeps running its old `Exec` until the next login, so the next agent applying a launcher change by hand (CLAUDE.md rule 9) plans for a logout instead of hunting for a live path.

- `.claude/skills/kde-config/lessons/shortcuts.md`: new section "A bound launcher's `Exec` is read once per login". The `KServiceActionComponent` keeps the `KService` it was built with; a ksycoca change only prunes components whose file vanished and creates ones for files carrying `X-KDE-Shortcuts`, so a launcher bound through `[services][<file>.desktop]` is never re-read. Includes the `busctl … invokeShortcut` call that shows what the daemon actually launches.
- `.claude/skills/kde-config/SKILL.md`: the Copilot-key row of the settings table now points at that lesson.

No provisioning change. The lesson deliberately recommends a logout over forcing a live component rebuild.

## Verification

- `pre-commit run --all-files` passes locally.
- Container `--check` runs in CI (the local Docker daemon is off).


.claude/skills/kde-config/lessons/shortcuts.md
```diff
@@ -68,6 +68,38 @@ alternative (`key[0].toCombined()`), and an unbound action as `[0]`.
 `Utils::normalizeSequences` drops empty sequences on write, so a leading
 zero must be filtered before re-sending the list.
 
+## A bound launcher's `Exec` is read once per login
+
+Confirmed live after the `Exec` of `hanzo-claude-code.desktop` was
```
palazzem: Uneeded. You are explaining the history. This is not a journal, remove all the prose and limit to write the rule, meaning how it works. This is one paragraph, not a storytelling.

palazzem: Cut to one paragraph: the component is built once at daemon start and reused on every `_launch`, a ksycoca change never re-reads a `[services]`-bound launcher, so an edited `Exec` applies at the next login. The `busctl … invokeShortcut` check stays as the last sentence.

— Claude

